Issue #30, opened 2026-05-26
Concurrent ListWorkouts/ListPrograms deadlock and exhaust the SQLite connection pool


Cawlumm:
## Severity
High — under concurrent load the backend can deadlock and stop serving entirely (connection-pool exhaustion). Pre-existing on `main`; surfaced while validating the #25 fix.

## Symptom (measured)
With the pool at `MaxOpenConns(10)`, firing concurrent `GET /workouts`:

| concurrency | result |
|---|---|
| 5 | ok, fast |
| **10** | **deadlock — all requests time out (~12s), pool stays exhausted afterward** |
| 20 | deadlock |

Once it deadlocks, the connections are held forever, so *every* subsequent request (any endpoint) also fails — the process is effectively down until restarted.

## Root cause
The load helpers issue a query **while a parent `rows` cursor is still open**, so a single list request needs more than one connection at once:

- `ListWorkouts` keeps its `rows` cursor open and calls `loadWorkoutExercises(w.ID)` inside the loop.
- `loadWorkoutExercises` keeps *its* cursor open and calls `loadSets(we.ID)` inside *its* loop.
- Same chain in `programs.go`: `ListPrograms` → `loadProgramExercises` → `loadProgramSets`.

With `MaxOpenConns(N)`, `N` concurrent list requests each grab one connection for their outer cursor, then all block waiting for a second connection for the nested query — which never frees. Classic pool-starvation deadlock. (This is also why `MaxOpenConns(1)` had to be backed out of the #25 fix — it dropped the trigger threshold to a single request; see commit on `fix/sqlite-locking`.)

Affected: `ListWorkouts`, `GetWorkout`, `CreateWorkout`/`UpdateWorkout` (post-commit reload), and the `programs.go` equivalents.

## Fix
Never query while a cursor is open: read the parent rows fully into a slice, `Close()` the cursor, then load children. e.g.

```go
rows, err := db.DB.Query(...)
// scan all parents into a slice
rows.Close()
for i := range parents { parents[i].Children = loadChildren(parents[i].ID) }
```

This falls out naturally from the per-entity store/"writer" classes planned in #29 (each store method finishes its query before the next runs), so the cleanest path is to fix it as part of that refactor. If #29 slips, the `load*` helpers can be restructured directly as a standalone fix.

Refs #25, #29.

Cawlumm:
[...]

Issue #29, opened 2026-05-26
Move DB access into per-entity store/writer classes (centralize SQL + fix nested-cursor deadlock)


Cawlumm:
## Context

Follow-up to the locking fix (#25, branch `fix/sqlite-locking`). That branch stays minimal: the corrected DSN (`busy_timeout`/WAL) fixes the lock, plus the `auth` register/login error mapping and the shared `utils.DBError` classifier (wired into `auth` only for now).

Today SQL is scattered directly across the HTTP handlers — **~61 `db.DB.*` call sites and ~73 inline SQL statements across 8 controllers**, each hand-writing its own error mapping. That duplication is what allowed the #25 bug class, and it also hides a serious latent bug (see #30).

## Goal

Move every entity's data access into its own **store / "writer" class** so SQL lives in one place per entity and the HTTP handlers stop touching the DB directly.

## Approach (from investigation)

A lightweight **per-entity store package** on plain `database/sql` (Navidrome-style; no ORM, no codegen for now):

- `store.UserStore`, `store.WeightStore`, `store.WorkoutStore`, `store.ProgramStore`, `store.FoodStore`, `store.ExerciseStore`, … each holding `*sql.DB` and exposing `List/Get/Create/Update/Delete` (+ a `Tx(func(*sql.Tx) error)` helper for the multi-statement workout/program creates).
- Controllers shrink to: parse → validate → call store → `utils.DBError` → respond.
- Stores stay HTTP-agnostic and return raw errors (`sql.ErrNoRows` + sqlite codes); the controller boundary maps them via `utils.DBError` / `IsLocked` / `IsUniqueViolation`.

## Must-fix deliverables folded into this work

1. **Fix the nested-cursor pool-exhaustion deadlock (#30).** `ListWorkouts`/`ListPrograms`/`loadWorkoutExercises`/`loadProgramExercises` currently query while a parent cursor is open, which deadlocks the pool under concurrency. Store methods must read a parent result fully (close the cursor) before loading children — which the writer-class structure enforces naturally.
2. **Finish the `DBError` rollout** that was reverted out of the #25 branch: the mistranslation fixes for `weight.go` (`GetWeightLog`), `food.go` (`GetFoodLog`), `exercises.go` (`GetExercise`), and the generic `InternalError`-on-lock sites across `user.go`, `active_session.go`, `programs.go`, `workouts.go`. Once SQL moves into stores, these route through one mapping path by construction.

## Alternatives considered
[...]

Issue #28, opened 2026-05-26
Enable SQLite foreign-key enforcement (cascade deletes) + clean up existing orphans


Cawlumm:
## Background

`backend/db/sqlite.go` originally tried to enable foreign keys via the DSN, but the param was written in `mattn`-style syntax (`_foreign_keys=on`) that the `modernc.org/sqlite` driver silently ignores — so **foreign-key enforcement has never actually been on**. (Verified: a connection opened with the old DSN reports `PRAGMA foreign_keys = 0`.)

The fix for #25 corrected the DSN syntax for `busy_timeout` + WAL but **deliberately left `foreign_keys` off**, because turning it on is a behavior change that deserves its own change + testing.

## Why this matters

The schema is full of `ON DELETE CASCADE` (`backend/db/migrations.go`): `workouts`, `sets`, `weight_logs`, `food_logs`, `saved_foods`, `programs`, etc. all reference `users(id) ON DELETE CASCADE`. With FK enforcement off:

- **Deleting an account (`DELETE FROM users`) does not cascade** → every workout/set/weight/food/program row for that user is orphaned and lingers forever.
- FK integrity isn't enforced at all (e.g. a row could reference a non-existent parent).

## Proposed work

1. Add `&_pragma=foreign_keys(on)` to the DSN in `backend/db/sqlite.go`.
2. **Audit delete paths** for anything that currently relies on FKs being off (e.g. manual deletes that would now fail or cascade unexpectedly), and confirm insertion ordering satisfies constraints.
3. **One-time orphan cleanup migration** for existing databases — delete rows whose parent no longer exists, since enabling enforcement does *not* retroactively remove pre-existing orphans. (Consider `PRAGMA foreign_key_check` to enumerate them.)
4. Tests: a regression test that deleting a user cascades to their child rows.

## Notes

- `foreign_keys` is a **per-connection** pragma, so it must live in the DSN (applied to every pooled connection) — which the corrected `_pragma` syntax now does.
- Split out from #25 so the locking fix could ship without bundling a data-affecting behavior change.

Cawlumm:
Substantially addressed by #57 (merged). Status against the proposed work:

1. ✅ DSN now sets `_pragma=foreign_keys(on)` — FK enforcement is on (modernc honors it).
[...]

Issue #25, opened 2026-05-24
sqlite locked DB results in incorrect errors


shaine:
**What happened?**
When the sqlite database is locked, the authentication errors are always incorrect. I discovered the DB was locked after I was already logged in and poking around. Logging out and back in consistently gave "incorrect username/password", and the registration flow consistently returned "email already exists".

As for why the database was locked... I don't know, sqlite mystifies me in this regard. I/O bottleneck on my side, perhaps. Restarting the backend server unlocked the database, seemingly. Would be nice if the server encountered locks and attempted to take steps to fix the lock.

**Steps to reproduce**
1. _waves hand_ lock the database somehow? 😹 
2. Attempt to log in or register
3. See error

**Expected behavior**
1. Attempt to unlock the database
2. At the end of unlock attempts, return a more appropriate error such as "Database could not be reached." or "Unable to perform {x task}, please contact the administrator."
3. Log details specific to the attempts and the lock

**Environment**
- Deployment: Docker / Synology NAS
- Browser: Desktop Firefox
- OS: Ubuntu 24

**Logs or screenshots**
[lyftr-backend] 2026-05-24T22:40:39.490904115Z [GIN] 2026/05/24 - 22:40:39 | 200 | 63.595µs |       127.0.0.1 | GET      "/health"
[lyftr-backend] 2026-05-24T22:40:45.401787123Z [GIN] 2026/05/24 - 22:40:45 | 204 | 69.533µs |    192.168.50.1 | OPTIONS  "/api/v1/auth/register"
[lyftr-backend] 2026-05-24T22:40:45.492162030Z [GIN] 2026/05/24 - 22:40:45 | 400 |  87.17ms |    192.168.50.1 | POST     "/api/v1/auth/register"
[lyftr-backend] 2026-05-24T22:41:10.321257524Z [GIN] 2026/05/24 - 22:41:10 | 200 | 52.977µs |       127.0.0.1 | GET      "/health"
[lyftr-backend] 2026-05-24T22:41:35.774490121Z [GIN] 2026/05/24 - 22:41:35 | 204 | 85.259µs |    192.168.50.1 | OPTIONS  "/api/v1/auth/login"
[lyftr-backend] 2026-05-24T22:41:35.776263643Z [GIN] 2026/05/24 - 22:41:35 | 401 | 305.817µs |    192.168.50.1 | POST     "/api/v1/auth/login"
[...]

PR #32, opened 2026-05-26
fix: resolve SQLite lock failures under concurrent writes

Cawlumm:
## Summary
Fixes the "database is locked" failures under concurrent use. Root cause: the `modernc.org/sqlite` connection string used mattn-style params (`_busy_timeout=…&_journal_mode=…`) that this driver **silently ignores**, so `busy_timeout` was effectively `0` and any contended write failed instantly.

- **`db/sqlite.go`** — switch to modernc's `_pragma(...)` DSN syntax: `busy_timeout(5000)` (wait out contention, including cross-process like a NAS backup, instead of erroring immediately), `journal_mode(WAL)` (readers don't block the writer), `synchronous(NORMAL)`. The pool stays at 10/5 — a single connection is **not** safe here (some handlers query while a parent cursor is open; see follow-up below).
- **`auth.go`** — register/login no longer mistranslate a DB error into the wrong domain message. Registration returns **409** only on a genuine `UNIQUE` violation and **503** (retryable) on a lock, instead of telling a brand-new user "email already registered".
- **`utils.DBError` + `IsLocked`/`IsUniqueViolation`** — a shared classifier so a locked DB becomes a logged, retryable 503. Wired into `auth` here; the broader rollout is a follow-up.

## Verification
- `go build ./...`, `go vet ./...`, `go test ./controllers/ ./utils/` — all green.
- New tests: `sqlite_lock_test.go` reproduces the issue (forced lock → 503; real `UNIQUE` → 409); `dberr_test.go` covers the classifier.
- Load check: **300 concurrent writes → 300/300 succeed** on this branch vs **21/300 on `main`**.
- UI before/after through the real register form under a write-storm: `main` shows "email already registered" on a brand-new email; this branch registers successfully and lands on the dashboard.

## Scope / follow-ups (out of scope of this fix — tracked separately)
- **#29** — move DB access into per-entity store / "writer" classes (centralize SQL) and finish the `DBError` rollout across the remaining controllers.
- **#30** — pre-existing nested-cursor deadlock in `ListWorkouts`/`ListPrograms` (also affects `main`); resolved naturally by #29's restructure.
- **#28** — enable SQLite foreign-key enforcement + orphan cleanup.

These are intentionally left to follow-up branches to keep this fix minimal and reviewable.

## Issue
Addresses #25. Leaving the issue **open** on purpose — @Cawlumm (the reporter), please validate the fix and close it yourself.

https://claude.ai/code/session_01EeGweBeouxSsT6nSaXy8Pk

---
_Generated by [Claude Code](https://claude.ai/code/session_01EeGweBeouxSsT6nSaXy8Pk)_

### DIFF
--- a/backend/controllers/active_session.go
+++ b/backend/controllers/active_session.go
@@ -21,7 +21,8 @@ func GetActiveSession(c *gin.Context) {
 		utils.OK(c, nil)
 		return
 	}
-	if utils.DBError(c, err) {
+	if err != nil {
+		utils.InternalError(c)
 		return
 	}
 	c.JSON(200, gin.H{"data": gin.H{"data": data, "updated_at": updatedAt}})
@@ -42,7 +43,8 @@ func UpsertActiveSession(c *gin.Context) {
 		 ON CONFLICT(user_id) DO UPDATE SET data = excluded.data, updated_at = CURRENT_TIMESTAMP`,
 		uid, body.Data,
 	)
-	if utils.DBError(c, err) {
+	if err != nil {
+		utils.InternalError(c)
 		return
 	}
 	utils.OK(c, gin.H{"saved": true})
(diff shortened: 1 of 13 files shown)


Cawlumm:

PR #36, opened 2026-06-10
fix: never run nested queries while a rows cursor is open

### LINKED ISSUE #30: Concurrent ListWorkouts/ListPrograms deadlock and exhaust the SQLite connection pool

Cawlumm:
## Summary
Fixes the nested-cursor connection-pool deadlock from #30. `ListWorkouts`/`GetWorkout`/create-update reloads and the `programs.go` equivalents issued child queries (`loadWorkoutExercises` → `loadSets`) while the parent `rows` cursor was still open, so a single request needed up to three pool connections at once. With `MaxOpenConns(10)`, ten concurrent list requests deadlocked the pool permanently — every subsequent request on any endpoint failed until restart.

## Fix
Scan all parent rows into a slice and close the cursor **before** loading children, in both the handlers and the `load*` helpers — one connection per request now suffices. Also added `Scan`/`rows.Err` checks that were previously ignored (failed scans silently appended zero-valued rows).

## Verification
- New regression tests (`concurrency_test.go`) cap the pool at 2 and fire 6 concurrent list requests: **deadlock + 10s timeout on `main`, pass on this branch** (verified both ways).
- Live check: 20 concurrent `GET /api/v1/workouts` against a running backend → 20/20 × 200 OK (the #30 table shows total deadlock at 10 on `main`).
- `go build ./...`, `go vet ./...`, `go test ./controllers/` green; full Playwright e2e suite: 161 passed, 2 skipped.

Closes #30.

### DIFF
--- a/backend/controllers/concurrency_test.go
+++ b/backend/controllers/concurrency_test.go
@@ -1,0 +1,124 @@
+package controllers
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/Cawlumm/lyftr-backend/db"
+)
+
+// Regression test for #30: list handlers used to run nested queries while a
+// parent rows cursor was still open, so each request needed multiple pool
+// connections at once. With the pool capped at N, N concurrent requests
+// deadlocked the pool permanently. The fix scans parents fully and closes the
+// cursor before loading children, so one connection per request suffices.
+func TestListWorkoutsConcurrentDoesNotExhaustPool(t *testing.T) {
+	setupTestDB(t)
+	uid := createTestUser(t)
+	exID := createTestExercise(t)
+
+	// Two workouts with exercises and sets so the nested loaders actually run.
+	for w := 0; w < 2; w++ {
+		res, err := db.DB.Exec(
+			`INSERT INTO workouts (user_id, name) VALUES (?, ?)`, uid, fmt.Sprintf("Workout %d", w),
+		)
+		if err != nil {
+			t.Fatalf("insert workout: %v", err)
+		}
+		wid, _ := res.LastInsertId()
+		weRes, err := db.DB.Exec(
+			`INSERT INTO workout_exercises (workout_id, exercise_id, order_index) VALUES (?, ?, 0)`, wid, exID,
+		)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


Cawlumm:

PR #38, opened 2026-06-10
fix: claim seeding flag atomically before wiping exercises

Cawlumm:
## Summary
`WipeAndReseed` checked `seeding.Load()` but the flag was only set inside the async goroutine (`fetchAndStore`), leaving a window — from the check until the goroutine got scheduled — where the flag was still `false`. Two rapid `POST /admin/reset-exercises` calls could both pass the check, wipe twice (the second wipe deleting mid-seed rows), and run two seeds concurrently. `SyncExercises` had no concurrency guard at all.

## Fix
- `WipeAndReseed` claims the flag with `CompareAndSwap` **before** the wipe and holds it until the background sync finishes (released on wipe failure).
- `fetchAndStore` is now CAS-guarded too, so the admin sync endpoint and startup seed can't overlap a running seed; the actual sync lives in `fetchAndStoreLocked`.

## Verification
`go build ./...`, `go vet ./...`, `go test ./controllers/` green; Playwright e2e: 161 passed, 2 skipped.

### DIFF
--- a/backend/seed/exercises.go
+++ b/backend/seed/exercises.go
@@ -72,21 +72,36 @@ func SyncExercises(db *sql.DB) error {
 
 // WipeAndReseed deletes all exercises then re-fetches from source.
 func WipeAndReseed(db *sql.DB) error {
-	if seeding.Load() {
+	// Claim the flag atomically BEFORE wiping: a Load-then-act check races —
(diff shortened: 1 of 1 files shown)


Cawlumm:

PR #57, opened 2026-06-29
refactor: move all SQL into per-entity store classes (repository pattern)

### LINKED ISSUE #29: Move DB access into per-entity store/writer classes (centralize SQL + fix nested-cursor deadlock)

Cawlumm:
## What & why

Controllers were building and running SQL inline (114 `db.DB.` calls across 8 controllers). This moves **all** of it into a per-entity **store layer** (repository pattern): only stores touch the database; controllers do HTTP only (bind → validate → call store → map result/error). Since we were already in this code, this **absorbs three in-flight fixes** into one MR instead of merging them piecemeal.

**Supersedes #32, #36, #38. Closes #29.** (Also fixes the underlying #25 production 500 and #30 pool deadlock.)

## Architecture (full constructor injection)

- New `backend/stores` package — one store per entity (`Weight`/`Food`/`User`/`Exercise`/`Workout`/`Program`/`ActiveSession`), aggregated by `stores.Stores` + `New(db)`. Stores own all SQL + transactions, return models + raw errors, and import neither `gin` nor `utils`.
- Controllers are methods on `controllers.Handler{ s *stores.Stores }`; `routes.Setup(r, h)` injects it; `main` wires `stores.New(db.DB) → NewHandler`.
- `utils.DBError(c, err)` is the standard controller tail (lock → 503 retryable, else → 500); unique violations → 409.

## Absorbed fixes

- **#32 (prod 500):** the mattn-style DSN params were silently ignored by `modernc.org/sqlite`, leaving `busy_timeout` at 0 → contended writes 500'd instantly. DSN now uses `_pragma=…` (busy_timeout + WAL + synchronous + **foreign_keys(on)** — #32 had accidentally dropped FK). Adds `utils/dberr.go` + `Conflict`/`ServiceUnavailable`.
- **#36 (pool deadlock):** list loaders now scan + close the parent cursor **before** loading children, so a request never needs two pool connections; scan errors are surfaced, not swallowed.
- **#38 (reseed race):** atomic `CompareAndSwap` seed guard.

## Working *with* SQLite, not against it

SQLite is a single writer. Rather than fight that with a larger pool (which just turns contention into lock errors to retry), `MaxOpenConns(1)` serializes all DB access through one connection — safe now that #36 means no request needs a second connection mid-query, so there's no in-process lock contention to fail on. WAL/busy_timeout remain only as cheap cross-process defense.

## Verification

- `go build ./... && go test ./...` green at **every commit** (11 commits, each compiles + passes).
- Folds in the #32 lock tests (503-on-lock, 409-on-duplicate) and #36 concurrency tests (pool=2, 6 goroutines, deadlock guard) — adapted to DI.
- New `TestDeleteAccountCascadesChildren` guards the FK-cascade behavior — and **exposed a latent test bug**: the harness DSN also used the ignored `_foreign_keys=on`, so the suite had been running with FK enforcement *off*. Fixed to `_pragma=foreign_keys(on)` so tests match production.
- **Playwright e2e: 87 passed, 1 skipped** against the rebuilt backend (chromium + mobile), all test accounts swept clean.
- `grep db.DB. backend/controllers/*.go` → **zero** live references.

### DIFF
--- a/backend/controllers/active_session.go
+++ b/backend/controllers/active_session.go
@@ -2,9 +2,7 @@ package controllers
 
 import (
 	"database/sql"
-	"time"
 
-	"github.com/Cawlumm/lyftr-backend/db"
 	"github.com/Cawlumm/lyftr-backend/middleware"
 	"github.com/Cawlumm/lyftr-backend/utils"
 	"github.com/gin-gonic/gin"
@@ -12,17 +10,12 @@ import (
 
 func (h *Handler) GetActiveSession(c *gin.Context) {
 	uid := middleware.UserID(c)
-	var data string
-	var updatedAt time.Time
-	err := db.DB.QueryRow(
-		`SELECT data, updated_at FROM active_sessions WHERE user_id = ?`, uid,
-	).Scan(&data, &updatedAt)
+	data, updatedAt, err := h.s.ActiveSession.Get(uid)
 	if err == sql.ErrNoRows {
 		utils.OK(c, nil)
 		return
 	}
-	if err != nil {
-		utils.InternalError(c)
+	if utils.DBError(c, err) {
 		return
 	}
 	c.JSON(200, gin.H{"data": gin.H{"data": data, "updated_at": updatedAt}})
@@ -37,14 +30,7 @@ func (h *Handler) UpsertActiveSession(c *gin.Context) {
 		utils.BadRequest(c, err.Error())
 		return
 	}
-	_, err := db.DB.Exec(
-		`INSERT INTO active_sessions (user_id, data, updated_at)
-		 VALUES (?, ?, CURRENT_TIMESTAMP)
-		 ON CONFLICT(user_id) DO UPDATE SET data = excluded.data, updated_at = CURRENT_TIMESTAMP`,
-		uid, body.Data,
-	)
-	if err != nil {
-		utils.InternalError(c)
+	if utils.DBError(c, h.s.ActiveSession.Upsert(uid, body.Data)) {
 		return
 	}
 	utils.OK(c, gin.H{"saved": true})
@@ -52,6 +38,8 @@ func (h *Handler) UpsertActiveSession(c *gin.Context) {
 
 func (h *Handler) DeleteActiveSession(c *gin.Context) {
 	uid := middleware.UserID(c)
-	db.DB.Exec(`DELETE FROM active_sessions WHERE user_id = ?`, uid)
+	if utils.DBError(c, h.s.ActiveSession.Delete(uid)) {
+		return
+	}
 	utils.OK(c, gin.H{"deleted": true})
 }
--- a/backend/controllers/active_session.go
+++ b/backend/controllers/active_session.go
@@ -10,7 +10,7 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-func GetActiveSession(c *gin.Context) {
+func (h *Handler) GetActiveSession(c *gin.Context) {
 	uid := middleware.UserID(c)
 	var data string
 	var updatedAt time.Time
@@ -28,7 +28,7 @@ func GetActiveSession(c *gin.Context) {
 	c.JSON(200, gin.H{"data": gin.H{"data": data, "updated_at": updatedAt}})
 }
 
-func UpsertActiveSession(c *gin.Context) {
+func (h *Handler) UpsertActiveSession(c *gin.Context) {
 	uid := middleware.UserID(c)
 	var body struct {
 		Data string `json:"data" binding:"required"`
@@ -50,7 +50,7 @@ func UpsertActiveSession(c *gin.Context) {
 	utils.OK(c, gin.H{"saved": true})
 }
 
-func DeleteActiveSession(c *gin.Context) {
+func (h *Handler) DeleteActiveSession(c *gin.Context) {
 	uid := middleware.UserID(c)
 	db.DB.Exec(`DELETE FROM active_sessions WHERE user_id = ?`, uid)
 	utils.OK(c, gin.H{"deleted": true})
--- a/backend/controllers/auth.go
+++ b/backend/controllers/auth.go
@@ -3,7 +3,6 @@ package controllers
 import (
 	"database/sql"
 
-	"github.com/Cawlumm/lyftr-backend/db"
 	"github.com/Cawlumm/lyftr-backend/models"
 	"github.com/Cawlumm/lyftr-backend/utils"
 	"github.com/gin-gonic/gin"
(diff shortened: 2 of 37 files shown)


Cawlumm:
## 🔍 Code review (workflow-backed, high effort)

8 finder angles → 35 candidates → independent verify pass → **10 findings kept, 10 refuted/dropped**. Ranked by severity, with disposition. Most stem from this PR *actually enabling FK enforcement* (the old `_foreign_keys=on` was silently ignored) and the `MaxOpenConns(1)` serialization.

### 🔴 Will fix — correctness regressions
1. **`seed/exercises.go` — `WipeAndReseed` 400s on any live DB.** With FK now enforced, `DELETE FROM exercises` is rejected (RESTRICT) once any workout/program references an exercise → admin reset-exercises breaks. Fix: prune only *unreferenced* exercises; the reseed already upserts via `ON CONFLICT(name)`, so referenced rows get refreshed in place.
2. **`stores/weight.go:87` — `UpsertForDay` is a non-transactional read-modify-write.** The connection is released between the existence `SELECT` and the `INSERT`, so two same-day logs can both insert (duplicate day). Fix: wrap in a transaction (with `MaxOpenConns(1)` the tx holds the one connection → the RMW is atomic).
3. **`stores/exercise.go:73 / :27` — `Get` swallows the scan error.** `scanExercise` discards `Scan`'s error and `Get` infers not-found from `id==0`, so a transient lock/scan failure is reported as a 404, not the 503/500 the dberr layer exists to send. Fix: capture the error; only real `sql.ErrNoRows` → not found.
4. **`controllers/user.go:15` — `GetMe` ghost account.** `DBError` ignores `ErrNoRows`, so a deleted-but-valid-JWT user gets `200` with an empty user. Fix: `ErrNoRows` → `401`.
5. **`stores/workout.go:170` (+ programs) — bad `exercise_id` now 500s.** FK rejects a workout/program referencing a non-existent exercise (previously a silent dangling insert). Fix: map FK violation → `400` with a clear message (new `IsForeignKeyViolation` classifier).

### 🟠 Will fix — error-swallowing / consistency
6. **`stores/weight.go:177` — `changeOver` discards its query error and returns `0`,** so `GetWeightStats` can show a fake `0.0` change instead of a 503/500. Fix: propagate the error.
7. **`controllers/food.go` (×6) — hand-rolled `c.JSON(503,…)`** in the OFF proxy bypass the new `utils.ServiceUnavailable` (no `Retry-After`). Fix: use the helper.

### 🟡 Flagged — deliberate trade-offs / verified OK (not changing here)
8. **`db/sqlite.go:43` — `MaxOpenConns(1)` lets a long admin `sync-exercises` block all requests.** This is the intended "serialize on one connection" decision; sync is admin-only + rare. Mitigation (follow-up): commit the import in batches so it yields the connection. *Leaving as-is — it's the chosen model.*
9. **`stores/workout.go:60` / `program.go:59` — N+1 list loaders.** Real, but pre-existing and **deliberately deferred** in the plan (close-cursor-then-load was chosen over set-based `WHERE IN` to limit this PR's risk). Follow-up: batch with `WHERE parent_id IN (...)`.
10. **`auth.go:33` — duplicate-register 400→409.** **Verified OK:** `Register.tsx` renders the server's error message regardless of status, so "email already registered" still shows; 409 is the more correct status. No change.

Addressing 1–7 now; pushing fixes to this branch.

Cawlumm:
## ✅ Addressed in `99961ae`

Fixed findings **1–7** (all build + full suite green, 2 new regression tests added):

| # | Finding | Fix |
|---|---|---|
| 1 | `WipeAndReseed` 400s under FK | prune only *unreferenced* exercises; upsert refreshes the rest |
| 2 | `UpsertForDay` non-transactional RMW | wrapped the check + write in a transaction |
| 3 | `ExerciseStore.Get` masks errors as 404 | `scanExercise` returns the scan error; only real `ErrNoRows` → not found |
| 4 | `GetMe` ghost account | `ErrNoRows` → `401` |
| 5 | unknown `exercise_id` → 500 | new `IsForeignKeyViolation` → `400` on workout/program create+update |
| 6 | `changeOver` swallows error | propagates it through `Stats` |
| 7 | food-proxy hand-rolled 503s | use `utils.ServiceUnavailable` (Retry-After) |

New tests: `TestGetMe_deletedAccountReturns401`, `TestCreateWorkout_unknownExerciseReturns400`.

### Intentionally not changed (with rationale)
- **#8 `MaxOpenConns(1)` + admin sync blocking** — this is the deliberate "serialize on one connection" model for SQLite's single writer; `sync-exercises` is admin-only and rare. Batched-commit mitigation noted as a future follow-up.
- **#9 N+1 list loaders** — pre-existing and deliberately deferred in the plan (close-cursor-then-load was chosen over set-based `WHERE IN` to bound this PR's risk). Good follow-up issue.
- **#10 (auth 400→409)** — verified no-op: the client renders the server error message regardless of status, and 409 is the more correct status.

Issue #81, opened 2026-08-03
Trust-on-first-use certificate approval (per-certificate, instead of trusting the whole user CA store)


Cawlumm:
Follow-up to #79.

#79 shipped an Android network security config with `<certificates src="user"/>`, so the app trusts CAs the device owner installed. Chain, expiry and hostname are still fully verified — it is not an ignore-TLS switch — and it solved the reported problem.

But the trust it grants is **app-wide over the whole user CA store**. If you installed a CA years ago for mitmproxy, a corporate MDM, or some other tool, that CA can now also intercept Lyftr traffic. Nothing about it is scoped to your server.

### Proposal

Trust-on-first-use, the model Nextcloud's Android client uses. On an untrusted-certificate failure, show the certificate — subject, issuer, validity, SHA-256 fingerprint — and let the user approve **that specific certificate**, pinned into an app-local keystore. Nothing else gains trust.

This is strictly better than what ships today: informed consent at the moment of failure, and per-certificate rather than per-CA-store scope.

### Why it wasn't in #79

It needs native code on both platforms — a custom `X509TrustManager` installed via `OkHttpClientProvider.setOkHttpClientFactory` on Android, and an `NSURLSession` delegate on iOS — plus the approval UI and a persistent keystore. The config-only fix in #79 was ~40 lines and unblocked the reporter completely, so this became a follow-up rather than a blocker.

Worth noting Nextcloud's own version of this dialog has a history of crashes ([#10871](https://github.com/nextcloud/android/issues/10871), [#11677](https://github.com/nextcloud/android/issues/11677)) — the design is right, the implementation needs care.

### Notes

- If this lands, `<certificates src="user"/>` could be dropped, or kept as a fallback.
- The genuinely best answer for most users remains a real certificate. Caddy can issue a public Let's Encrypt cert for a LAN-only hostname over DNS-01 with no ports exposed — now documented on the HTTPS page.
- Suggested by @pdschneider in #79, who asked for an "Accept Self-Signed Certificates" checkbox. A checkbox is the wrong mechanism (it would gate something the OS already gates, or else mean genuinely ignoring TLS errors), but the underlying instinct — explicit, scoped, informed consent — is right, and this is what it looks like built properly.

Issue #67, opened 2026-07-10
Dedup: web → @lyftr/shared (contract-only: types + pure utils)


Cawlumm:
Make the **web** app consume `@lyftr/shared` (canonical, already on `main` via mobile) instead of its own duplicated domain types + pure utils. Kills the triple-declared API-contract coupling (backend → web → mobile) with **zero app-behavior change**.

**Scope: contract-only.** Types + pure utils. DEFER the API client (singleton→factory) and Zustand stores (sync→async factories) — those are a behavioral refactor, separate pass.

### The real work is plumbing (4 resolution surfaces), not the type-swap
1. **Vite dev + Vitest** — web joins the root npm workspace → `@lyftr/shared` symlinked in. ⚠️ shared's entry is raw `src/index.ts` (no build step) — verify Vite/esbuild transpiles the linked workspace pkg.
2. **CI `unit` + `e2e`** (ci.yml) — `cd web && npm ci` → root `npm ci`; cache key `web/package-lock.json` → `package-lock.json`.
3. **`web/Dockerfile`** (dockerhub image + e2e `buildx bake`) — context is `./web`, can't see `packages/shared`. Broaden compose build context to repo root + rewrite for a workspace install.
4. **`fly/Dockerfile`** (PRODUCTION deploy) — context already root; add COPY of root `package.json`/lock + `packages/shared`, workspace `npm ci`, build web. ⚠️ next `main` deploy exercises it.

### Phases
- **Phase 0 — Wiring** (all risk lives here; no behavior change): root workspace membership, drop `web/package-lock.json`, fix Vite raw-TS resolution, rewrite both Dockerfiles + compose context, edit ci.yml install/cache. Gate: local `docker compose build frontend` + dev/build/test green.
- **Phase 1 — Types**: delete `web/src/types.ts`, import from `@lyftr/shared`. Free fix: web's `ActiveSessionSet` gains the missing `program_set_id` (#40 field).
- **Phase 2 — Pure utils**: weight-conversion fns (`web/src/stores/settings.ts`) + date utils (`web/src/utils/dateUtils.ts`) → shared; stores themselves stay in web.

### Logistics
Fresh worktree off `main`: `git worktree add ../repo-shared -b feat/web-shared-dedup main` (has web + packages/shared + fly together). One reviewable PR. Safety net = Playwright e2e (rebuilds both images) + local `docker compose build`. Do NOT add a new required status check.

[...]

PR #82, opened 2026-08-03
fix(mobile): trust user-installed CAs and allow cleartext to local servers (#79)

### LINKED ISSUE #79: Accept Self-Signed Certs
Hi, I just installed the Lyftr server and the Android mobile app. However, I keep getting this error when I try to reach my server:

"Can't reach the server. Check the URL, that the backend is running, and that it allows this app's origin (CORS)."

That happens with my apache reverse proxy custom .lan domain, the raw reverse proxy IP:Port, and the original non-reverse proxy IP:Port combo.

I have CORS_ORIGIN=* in my .env file.

According to the logs, my phone *is* reaching the server, but it is being denied. I'm not 100% sure, but my suspicion is that it is blocking my endpoint due to the certificate being self-signed and installed rather than being the system root CA.

If you are able to, please add an option to ignore TLS  / certificate errors. If you need any more information from me (logs,etc) please let me know.

Thanks!

Cawlumm:
Fixes #79.

## The problem

The Android app took Android's networking defaults, which block two separate things a self-hoster needs:

- **User-installed CAs are ignored** since API 24. A private CA installed on the device works in Chrome but not in the app — which is why the reporter's server looked fine in their phone's browser and failed in Lyftr.
- **Cleartext HTTP is refused** since API 28.

Then `apiErrorMessage` collapsed both into one message blaming CORS — which cannot apply, since a native app sends no `Origin` header. The reporter spent their time auditing `CORS_ORIGIN` for a block that was entirely client-side.

**The cleartext half was the wider bug.** `docker-compose.yml` defaults to plain HTTP, and `site/mobile.md` told users to enter `http://192.168.1.10:8080` — a URL the shipped APK was incapable of opening. That path was broken for every Android user on a default install, not just self-signed setups. #79 is the first report, not the first occurrence.

## The fix

**Android network security config**, via a local Expo config plugin (`mobile/plugins/`, no new dependency):

```xml
<base-config cleartextTrafficPermitted="true">
  <trust-anchors>
    <certificates src="system" />
    <certificates src="user" />
  </trust-anchors>
</base-config>
```

Chain, expiry and hostname are still fully verified. This **tightens** trust to the device owner's own store rather than disabling validation, and Android already gates that store behind a full-screen "your data won't be private" interstitial.

A blanket ignore-TLS toggle — what the issue actually asked for — was rejected on evidence: Immich ships one and it has been half-broken for over a year ([#15188](https://github.com/immich-app/immich/issues/15188), [#25692](https://github.com/immich-app/immich/issues/25692)), and Google publishes a [policy page](https://support.google.com/faqs/answer/6346016) against accept-all trust managers.

**A real error classifier** (`packages/shared/src/utils/networkError.ts`). Axios reports every transport failure as `'Network Error'` with code `ERR_NETWORK`, so the actual reason has to come off the XHR: React Native stores the native string in `_response` and wipes the `response` getter once `_hasError` is set. Unmatched strings are appended verbatim rather than guessed at — precisely the failure mode that rotted Immich's toggle.

**Two supporting fixes.** The main axios instance had no `timeout` (default `0` = wait forever), so a policy-blocked request hung the UI instead of failing. And the login screen's server row was probe-gated, so a user whose server is unreachable could never save a URL from it — the exact dead end in this issue. Now warn-but-save, matching Settings.

## Verification

On an Android 14 emulator, against nginx with a private CA:

| Case | Result |
|---|---|
| `http://10.0.2.2:3000` — the broken default-install path | ✅ connects |
| `https://10.0.2.2:8443`, CA **not** installed | ✅ **still rejected**, message names the certificate |
| `https://10.0.2.2:8443`, CA in the user store | ✅ connects |

The middle row is the one that matters — it proves trust was tightened, not disabled.

**Confirmed by the reporter** against their own Apache reverse proxy and self-signed CA: *"Connection successful, login credentials accepted."*

Suite: Go `controllers` + `routes` pass, `go vet` clean, shared 72/72 (22 new classifier tests), mobile + web type-checks clean, Playwright 88 passed / 1 skipped / 0 failed.

## Notes

- **Native change — ships only in a new EAS build, never via OTA.** Version is **0.3.0**, not 0.2.0: `mobile-v0.2.0` was already tagged and released on 2026-07-23 and contains none of this.
- Docs now cover the DNS-01 path for LAN-only hostnames (a genuine Let's Encrypt cert with no ports exposed), which is the best answer for anyone who owns a domain, plus the private-CA rules that actually bite — SAN mandatory, IP servers need an `IP Address` SAN entry.
- `web/src/services/api.ts` deliberately left alone: web has no `@lyftr/shared` dependency yet (#67), and on web the CORS wording is *correct*.
- Per-certificate TOFU trust — stronger than trusting the whole user CA store — is tracked in #81.


mobile/plugins/withNetworkSecurityConfig.js
```diff
@@ -0,0 +1,71 @@
+const fs = require('fs')
```
Cawlumm: is a custom plugin mobile/plugins/withNetworkSecurityConfig.js this best practice when it comes to this logic?

mobile/plugins/withNetworkSecurityConfig.js
```diff
@@ -0,0 +1,71 @@
...
+    app.$['android:networkSecurityConfig'] = `@xml/${RES_NAME}`
+    return cfg
+  })
+
+const withNetworkSecurityConfig = (config) => withManifestAttribute(withConfigFile(config))
+
+module.exports = createRunOncePlugin(
+  withNetworkSecurityConfig,
+  'lyftr-network-security-config',
+  '1.0.0',
```
Cawlumm: Is it okay to hardcode aversion for this plugin?

packages/shared/src/client.ts
```diff
@@ -70,8 +73,11 @@ export function createClient(storage: StorageAdapter, opts: ClientOptions = {})
     return apiUrl(base)
   }
 
+  // Without an explicit timeout axios waits forever (default 0), so a silently dropped
+  // or policy-blocked request hangs the UI instead of surfacing an error.
   const api: AxiosInstance = axios.create({
     headers: { 'Content-Type': 'application/json' },
+    timeout: 20000,
```
Cawlumm, line 87: Is there a cleaner way to handle this then just a basic timeout, research what the best practice is. If it is a timeout then make sure 20000 is the right time to set as a default

PR #83, opened 2026-08-05
Fix app-wide timezone handling for workouts, food queries, and server defaults

### LINKED ISSUE #2: App-wide timezone correctness: workouts, food daily queries, server defaults
Surfaced while hardening the weight feature. The weight code now uses a consistent local-time-write / UTC-store pattern (see `web/src/utils/dateUtils.ts` and the `±12h` padded `?from=&to=` filter in `backend/controllers/weight.go`), but the rest of the app has the bugs below.

### 1. AddWorkout date picker stores the wrong day for non-UTC users

`web/src/pages/AddWorkout.tsx:82`
```ts
started_at: new Date(formData.date).toISOString()
```
`new Date('2026-04-25')` parses as **UTC midnight**, not local. After save, a user in UTC-7 picking "April 25" sees the workout listed under "April 24" because the stored UTC instant is `2026-04-25T00:00:00Z`, which is `2026-04-24T17:00:00-07` when rendered locally.

**Fix:** use `dayToIsoNoon(formData.date)` from `web/src/utils/dateUtils.ts` (already added on this branch).

### 2. Food daily queries bucket by UTC date

`backend/controllers/food.go:26` and `:121`
```go
WHERE user_id = ? AND date(logged_at) = ?
```
SQLite's `date()` extracts the date from the UTC instant. A user in UTC-7 logging dinner at 8pm local stores `03:00:00Z` the next day, so it ends up under tomorrow's date in `GET /food` and the daily macro totals are wrong.

`backend/controllers/food.go:126` has the same problem for the workout-count-per-day stat:
```go
SELECT COUNT(*) FROM workouts WHERE user_id = ? AND date(started_at) = ?
```

**Fix:** mirror the weight pattern — accept `?from=&to=` from the client (client computes the local-day boundary) instead of `date(logged_at) = ?`. Or, alternatively, pass the user's TZ offset and apply it inside the query.

### 3. Server `time.Now()` defaults are server-local

Several handlers fall back to `time.Now()` when the frontend omits a timestamp:

- `backend/controllers/food.go:20,59,112` (`time.Now().Format("2006-01-02")` and `time.Now()`)
- `backend/controllers/workouts.go:86` (`req.StartedAt = time.Now()`)

The instant is correct, but the serialized format depends on the container's `TZ` env var. Inconsistent across deployments.

**Fix:** `time.Now().UTC()` everywhere (already applied to `weight.go` on this branch as the reference).

### 4. `Dashboard.tsx` weekly query uses local TZ

`web/src/pages/Dashboard.tsx:24` — `TODAY = new Date()`, then `format(TODAY, 'yyyy-MM-dd')` is passed to `/food/stats?date=...`. Since the backend buckets by UTC date (#2), the dashboard "today's calories" can drift by a day for non-UTC users.

Resolved automatically once #2 is fixed.

## Not actually bugs (verified)

- `weight.go` daily-cutoff math (`time.Now().AddDate(0, 0, -days)`): comparing instants, TZ-independent ✓
- Frontend display via `format(new Date(iso), '...')`: renders in browser local TZ ✓
- JSON wire format: Go's `time.Time` round-trips with offset preserved ✓

## Reference implementation

The weight feature on `feature/implement-weight-feature-OCAyJ` is the model:

- **Frontend writes** anchor at local noon and serialize to UTC ISO via `dayToIsoNoon()` in `web/src/utils/dateUtils.ts`
- **Backend** `time.Now().UTC()` for any default
- **Backend daily filtering** uses an explicit `?from=&to=` window with ±12h padding for `YYYY-MM-DD` and exact handling for RFC3339 — see `backend/controllers/weight.go` `parseDayOrTime()`
- **Frontend display** stays as `format(new Date(iso), ...)` — already correct app-wide

## Suggested branch / scope

`fix/timezone-app-wide` — three small, independent commits:
1. AddWorkout `dayToIsoNoon` fix + any other date-input writers (~30m)
2. Food daily query → range filter + frontend update to send `from`/`to` (~1h)
3. Sweep `time.Now()` → `time.Now().UTC()` in `food.go` and `workouts.go` (~10m)

Add regression tests mirroring `TestListWeightLogs_dateRangeWestTimezone` for any new range filter.

DisplacedForest:
Closes #2.

Three commits, matching the scope laid out in the issue. The weight feature's local-write / UTC-store pattern is the reference throughout. We hit this same class of bug in DisplacedForest/ha-hevy-tracker#3, so the shape was familiar.

**1. AddWorkout date picker** — `new Date('YYYY-MM-DD')` parses as UTC midnight, so west-of-UTC users saw their workout land on the previous day. Now uses `dayToIsoNoon()`. Also swapped the default/max date derivation from `toISOString().slice(0, 10)` (UTC date) to `todayStr()` (local date): the old max blocked "today" for users east of UTC in the evening.

**2. Food daily queries** — `GET /food` and `GET /food/stats` now accept `?from=&to=`. The web client sends exact RFC3339 local-midnight boundaries (new `localDayRange()` helper), so daily lists, macro totals, and the workout count all bucket by the user's local day. The legacy `date=` path is untouched as a fallback so older clients keep working.

One deliberate difference from the weight listing: no ±12h widening here. Padding works for weight because it's one noon-anchored entry per day, but on a single-day window it would double-count entries from adjacent days, which daily totals can't tolerate. Exact boundaries from the client instead. A bare `YYYY-MM-DD` in `from`/`to` means the UTC day, matching legacy semantics.

**3. Server defaults** — food's `time.Now()` fallbacks now go through the same `normalizeLoggedAt()` used by weight (and already by workouts), and `UpdateFoodLog` forces a client-supplied `logged_at` to UTC so a non-UTC offset can't break the stored-text ordering. The two `date=` fallback defaults are `time.Now().UTC()`.

**Tests:** added `TestListFoodLogs_localDayRangeWestTimezone` and `TestGetDailyStats_localDayRangeWestTimezone`, mirroring `TestListWeightLogs_dateRangeWestTimezone` (UTC-7 user, 8pm local dinner stored next-day UTC). Full backend suite passes. Web type-check, lint, and build pass; the 14 pre-existing unit failures in `workoutSession`/store tests fail identically on a clean `main` checkout, so they're unrelated.

Dashboard's "today" drift (#4 in the issue) resolves automatically via the stats change since it already passes the local date string.


Cawlumm:
Reviewed against #2; `go test ./controllers/ ./routes/` passes on the branch. Approach is right, and two things go beyond the issue in a good way: `max={todayStr()}` fixes a bug nobody named (east-of-UTC users couldn't pick "today" in the evening), and `localDayRange` using local-midnight → next-local-midnight instead of `+24h` gets DST days right at 23/25 hours. Skipping the ±12h widening on single-day windows was correct.

Three things before merge:

**1. Mobile still takes the legacy path.** `packages/shared/src/client.ts:183,188` sends only `date=`, so UTC bucketing survives there. Mobile writes food at local noon (`mobile/app/(tabs)/nutrition/log.tsx:222`), so a UTC+13 user's entry stores as `23:00Z` the prior day and vanishes from its own list. Mobile is most of our users, so "app-wide" isn't true yet.

Context you'll hit: `web/src/utils/dateUtils.ts` and `packages/shared/src/utils/dateUtils.ts` are two copies — identical code, different comments. Web can't import `@lyftr/shared` because `web` isn't in root workspaces. Minimal fix: add `localDayRange` to the shared copy too (with a test — shared's `dateUtils` has none) and spread it into `client.ts` as you did in `web/src/services/api.ts`.

**2. No `food_logs` counterpart to `normalizeWorkoutStartedAt`** (`backend/db/migrations.go:78`). It exists because `modernc.org/sqlite` writes `time.Time` as `t.String()` with the zone embedded, so text range comparison only holds while every row shares a suffix — and fixed-offset zones store as `+0800 +0800`, unreadable on the way back. Your write-side `.UTC()` covers new rows; rows written earlier by `time.Now()` under a non-UTC `TZ` aren't covered. Small blast radius (Docker defaults to UTC), but the comparison is now load-bearing for food.

**3. `UpdateFoodLog`** — `backend/controllers/food.go:155` bare `.UTC()`, and `stores/food.go:88` writes `logged_at` unconditionally, so a `PATCH` omitting it stores year 1 and the row drops out of every range query. Pre-existing, no client triggers it, but `normalizeLoggedAt` (zero → now) is what `LogFood` already uses and closes it for free.

Deliberately out of scope, both worth their own issues: `FoodStore.History` (`stores/food.go:140`) still buckets by `substr()`; pre-merge workouts stored at UTC midnight aren't backfilled, so west-of-UTC users will see new workouts right and old ones a day early.

**How do you want to split it?**

- **(a)** all three here
- **(b)** just #1 here so the fix reaches mobile; I'll open issues for #2 and #3
- **(c)** merge as-is, I'll take all three — nothing here is a regression, it just doesn't reach the phone

On the duplication: it's wider than `dateUtils` (types differ by 6 lines of 267, `programUtils` is the same logic twice, mobile keeps its own `workoutSets`/`number` with a `// keep in sync` header). That's #67 and it's ours, not yours — if you'd rather not work around the fork, (c) is genuinely fine.

Thanks for the PR writeup; the `ha-hevy-tracker` context and the note about grepping every extraction point were both useful.

DisplacedForest:
Hey, thanks for the thorough review. I'll go ahead and take all three here, makes sense to keep it in one place since it's all the same theme anyway.

Heads up though, I'm slammed at the moment so it might be later today or even later in the week before I can get to it. Didn't want you waiting on radio silence.

Will ping you when it's ready for another look.

Cawlumm:
No rush at all, take your time. Thanks for the contribution! Yeah that issue was created by me way earlier in the lifecyle of Lftyr. Now we have both web and mobile, so some more nuance. We'll be able to sort this out though and no rush on it! Sorry for the confusion!

DisplacedForest:
All good! I don't even remember how I came across your repo but when I saw a simple issue to help with thought I'd be a good samaritan lol

DisplacedForest:
Alright, got to all three. The shared client now sends the local-day range for food list/stats so mobile is off the legacy path, and localDayRange lives in the shared dateUtils with tests (TZ pinned in the jest config so the DST cases are deterministic). The migration got generalized into normalizeUTCTimestamps and runs for food_logs.logged_at too, under its own flag so already-migrated workout databases don't rescan. UpdateFoodLog uses normalizeLoggedAt now. Also added the db and routes packages to the backend CI test step since the migration tests weren't running there.

Ready for another look whenever you get a chance.

Issue #80, opened 2026-08-01
+- .1 lb Buttons


MikeZheng33:
**What problem does this solve?**
What are you trying to do that you can't do today?

Right now in the weight logging section, there are plus and minus buttons which go up and down .5 lbs but my weight doesn’t change that much every day. Could you add .1 buttons as well?

**Describe the feature**
What should it look like and how should it work?

There would be smaller increments for the + and - buttons in the weight screen

**Who benefits?**
Is this specific to your setup or broadly useful?

I think anyone using the weight logger would find this useful

**Anything else?**
Screenshots, mockups, links to similar implementations — anything that helps.

I attached a screenshot of the page

<img width="1290" height="2619" alt="Image" src="https://github.com/user-attachments/assets/30717a65-fd11-4a67-b924-f7947963d2b2" />

Cawlumm:
Shipped in #85.

The +/- buttons on bodyweight now move **0.1** instead of 0.5, on both web and
mobile — the weight page, the entry you tap into to edit, and the quick weigh-in
sheet on the dashboard.

Worth knowing alongside it: the field itself has always accepted typed 0.1 values
(that was #39, also yours), so for a bigger correction you can still type the number
rather than tapping to it. The buttons are for nudging.

Lifting weight in gym mode is unchanged at 2.5 — that one's a plate increment, not a
measurement.

Mobile needs a new build to pick this up; it'll be in the next release.

Thanks for the report, and for the screenshot — it made the fix obvious.

Issue #39, opened 2026-06-12
Exact Weight Inputs


MikeZheng33:
I like to log my exact weights every day. Im curious as to why lyftr requires .5 or .0 numbers. Is this a database limitation or intentional design? Could it be made to work with .1 precision?

MikeZheng33:
I looked into it a bit and in `web/src/pages/ActiveWorkout.tsx` line 358 has `step="0.5"` so I assume this was an intentional choice. Could you change this to 0.1?

Cawlumm:
made an adjustment, let me know if this solves your issue! @MikeZheng33

PR #92, opened 2026-08-07
refactor(weight): move web bodyweight inputs to StepperTile, drop WeightInput's stepper (#91)

### LINKED ISSUE #91: Converge web bodyweight inputs on StepperTile and drop WeightInput's stepper mode
The two platforms render the same control two different ways.

| | Bodyweight | Gym mode |
|---|---|---|
| Mobile | `StepperTile` | `StepperTile` |
| Web | `WeightInput` (buttons flanking the field) | `StepperTile` |

Web is the odd one out: it's the only place a stepper is `[−] [field] [+]` instead of
a tile with the value full-width and a split ⊖/⊕ footer below it.

That layout is the reason `StepperTile` exists. From its own comment:

> Value spans the full tile (buttons are below, not flanking) so long weights never clip.

Flanking buttons squeeze the field, which is what web still does on all four
bodyweight inputs.

## What to change

Move these to `StepperTile` + `NumberField`, matching how mobile already does it:

- `web/src/pages/Weight.tsx`
- `web/src/pages/WeightDetail.tsx`
- `web/src/components/EditWeightModal.tsx`
- `web/src/components/QuickWeighInSheet.tsx`

Nothing is lost in the move — mobile proves it:

- the unit goes in the tile label (`Weight (${wUnit})`, same as gym mode)
- the `max` clamp goes in the `onStep` handler
  (`clampStep(..., { max: maxWeight(unit) })`, exactly what the three mobile weight
  screens do today)

## The payoff is a deletion

Once those four are converted, **nothing on web renders `WeightInput`'s stepper**.
The remaining six call sites (`AddWorkout`, `EditWorkout`, `AddWorkoutModal`,
`EditWorkoutModal`, `ActiveWorkout`, `DayExercisesEditor`) all pass
`stepper={false}` already.

So this can delete from `WeightInput`:

- the `stepper` and `step` props
- both `+`/`−` buttons and their layout branch
- `STEP_DEFAULT`

leaving a bare number field, which is all those six callers ever wanted.
`BODYWEIGHT_STEP` stays and moves to the four call sites, mirroring mobile.

This also retires an awkward note added in #85: `WeightInput`'s `step` prop is
currently overridden by no caller at all.

## Verification

- 390px width on all four screens, no overflow
- `+`/`−` still steps 0.1 and clamps at 0 and `maxWeight(unit)`
- kg still steps 0.1 kg
- existing `weight.spec.ts` E2E still passes; `WeightInput.test.tsx` loses its
  stepper cases and gains them wherever the tile lives

Follow-up to #80 / #85.

Cawlumm:
Closes #91. Follow-up to #85, which named the stepper increments and exposed that the
two platforms render the same control two different ways.

## The inconsistency

| | Bodyweight | Gym mode |
|---|---|---|
| Mobile | `StepperTile` | `StepperTile` |
| Web | `WeightInput` (buttons flanking the field) | `StepperTile` |

Web was the only place a stepper was `[−] [field] [+]` rather than a tile with the
value full-width and a split ⊖/⊕ footer — the layout `StepperTile` exists for:

> Value spans the full tile (buttons are below, not flanking) so long weights never clip.

## The change

The four web bodyweight inputs — `Weight`, `WeightDetail`, `EditWeightModal`,
`QuickWeighInSheet` — now use `StepperTile` + `NumberField`, mirroring mobile: unit
in the tile label, `max` clamp in the `onStep` handler.

Their old `<label>Weight</label>` headings are gone, since the tile carries its own.

## The payoff is a deletion

Nothing renders `WeightInput`'s stepper any more, so it's gone: both buttons,
`stepper`, `step`, `STEP_DEFAULT`, and the size-dependent padding that existed only
to leave room for flanking buttons. The six remaining callers are sets-table rows
that all passed `stepper={false}` — they now get the bare field they always wanted.

## A regression this caught before it shipped

`NumberField` had no `step` attribute, so `type=number` fell back to `step=1` and the
browser rejected every tenth:

> Please enter a valid value. The two nearest valid values are 183 and 184.

That blocks submit natively, before any request is made. Bodyweight is the one field
that exists to hold tenths (#39, #80), so the conversion would have made the log form
useless for its actual purpose.

Decimal fields now declare 0.1; integer fields keep 1 — which also repairs the same
latent flaw in gym mode's weight tile, where it was invisible only because that tile
isn't inside a validated form.

**The suite could not have caught this:** both weight E2E cases filled integers,
which pass whether or not `step` is set. One now fills `170.4`. Verified it fails
without the fix and passes with it.

## Coverage moved rather than vanished

`clampStep` — the stepping math behind all seven steppers — had no direct test; it
was only exercised through `WeightInput`'s buttons, which this deletes. Now:

- `web/src/utils/number.test.ts` — rounding, float dust, both clamps, non-finite
  base, and a case locking `BODYWEIGHT_STEP` / `PLATE_STEP` / `REP_STEP` apart
- `web/src/components/ui/StepperTile.test.tsx` — button wiring, aria labels, the
  step-attribute regression

Net +12 web unit tests (133 → 145) despite removing four.

## Verification

- web `tsc` clean, **145 unit tests pass**
- **88 E2E pass**, 1 skipped
- Go `./controllers/` uncached pass (untouched, per repo policy)
- Manual at **390px** in a real browser: tile renders without overflow,
  183.6 → 183.7 → 183.6 with no float dust, `−` at 0 stays 0, and a decimal
  bodyweight now actually saves (screenshot evidence taken before and after the fix)

### DIFF
--- a/web/src/components/WeightInput.tsx
+++ b/web/src/components/WeightInput.tsx
@@ -1,116 +1,62 @@
-import { Minus, Plus } from 'lucide-react'
 import { useNumericText } from '../hooks/useNumericText'
-import { BODYWEIGHT_STEP, clampStep } from '../utils/number'
 
 interface Props {
   value: string
   onChange: (next: string) => void
   unit: string
-  /** Increment for the +/- stepper buttons. Typing is always 0.1-precision regardless. */
-  step?: number
   autoFocus?: boolean
   size?: 'sm' | 'md' | 'lg'
-  /** Show the +/- stepper buttons (prominent inputs). Off = bare field for compact rows. */
-  stepper?: boolean
   placeholder?: string
   disabled?: boolean
-  /** Optional upper bound (display units) — clamps the +/- stepper. Submit-time
-   *  validation/messaging is the caller's job (an html max would preempt it with
-   *  a native browser tooltip). */
-  max?: number
 }
 
-// Two different granularities that happen to be the same number right now, and must
-// not be collapsed into one: INPUT_STEP is what the field accepts typed (the #39
-// feature — always tenths, whatever the buttons do), STEP_DEFAULT is how far one
-// +/- tap moves. They coincide because bodyweight, the only caller that shows the
-// buttons, wants the finest step the field can hold (#80).
-//
-// Nothing overrides `step` today: the sets tables pass stepper={false}, and gym mode
-// builds its own tiles from StepperTile + NumberField (see PLATE_STEP). The prop
-// stays for a caller that wants a coarser jump.
-//
-// That gym mode uses a different component for the same job is the inconsistency
-// tracked in #91: web is the only place a stepper flanks the field instead of
-// sitting in a tile footer, and mobile's bodyweight screens already use the tile.
-// Converging the four web bodyweight inputs onto StepperTile deletes the stepper
-// half of this component outright — both buttons, `stepper`, `step`, STEP_DEFAULT —
-// since the remaining callers all pass stepper={false}. Deliberately a follow-up,
-// not part of #80: it is a visual change to four screens, not a constant.
+// Typing is always 0.1-precision (the #39 feature), whatever the caller does with
+// the value afterwards.
 const INPUT_STEP = 0.1
-const STEP_DEFAULT = BODYWEIGHT_STEP
 
-// Single component for every weight input in the app. Conversion-agnostic: the
-// caller owns lbs↔display unit (pass display-unit strings in/out). `stepper`
-// toggles the prominent +/- layout (bodyweight, gym mode) vs a bare field for
-// compact sets-table rows. (Gym set tiles use the borderless NumberField instead.)
+// A bordered number field with the unit rendered inside it, for the compact rows of
+// the sets tables (add/edit workout, program day editor, active workout).
+//
+// It used to carry an optional +/- stepper too, but every prominent input that
+// wanted buttons has moved to StepperTile — bodyweight in #91, gym mode long before
+// that — and the callers left here all opted out of it. Reach for StepperTile if you
+// need buttons; this component is the bare field on purpose.
 export default function WeightInput({
   value,
   onChange,
   unit,
-  step = STEP_DEFAULT,
   autoFocus = false,
   size = 'md',
-  stepper = true,
   placeholder = '0.0',
   disabled = false,
-  max,
 }: Props) {
   // Raw typed text (see useNumericText) so in-progress entry isn't clobbered by the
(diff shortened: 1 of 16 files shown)


web/src/components/WeightInput.tsx
```diff
@@ -1,116 +1,62 @@
...
         onKeyDown={e => { if (['e', 'E', '+', '-'].includes(e.key)) e.preventDefault() }}
         step={INPUT_STEP}
         min="0"
         autoFocus={autoFocus}
         disabled={disabled}
-        className={`input ${inputSize} ${pad} text-center w-full tabular-nums ${disabled ? 'opacity-40' : ''}`}
+        className={`input ${inputSize} pr-7 text-center w-full tabular-nums ${disabled ? 'opacity-40' : ''}`}
         placeholder={placeholder}
       />
-      <span className={`absolute ${unitPos} top-1/2 -translate-y-1/2 text-xs text-tx-muted pointer-events-none`}>{unit}</span>
```
Cawlumm, line 95: Why did we remove this code block

Cawlumm: That's the stepper render block — `if (!stepper) return field`, the button styling, and the two `<button>` elements. Removing it is the point of #91 rather than a side effect, so it's fair to ask.

Nothing renders it any more. There were ten `WeightInput` call sites:

- the **four** that showed buttons (Weight, WeightDetail, EditWeightModal, QuickWeighInSheet) moved to `StepperTile` in this PR
- the other **six** are sets-table rows that already passed `stepper={false}`, so they never reached this code

With no caller taking the branch, keeping it would leave two components rendering the same control two different ways — which is the inconsistency the issue was filed about. The comment I'd added in #85 admitted it was already unused: *"Nothing overrides `step` today."*

Behaviour isn't lost, it relocated:

| Was here | Now |
|---|---|
| `+`/`−` buttons, aria labels | `StepperTile` |
| `adjust()` → `clampStep` | the `onStep` handler at each of the four call sites |
| `max` clamping | same handlers, `clampStep(..., { max: maxWeight(unit) })` |

Tests moved with it, and got better: `clampStep` had **no direct test at all** — it was only covered through these buttons — so deleting them without replacement would have dropped the stepping math to zero coverage. It's now in `number.test.ts` (rounding, float dust, both clamps, non-finite base), with the button wiring in `StepperTile.test.tsx`.

What `WeightInput` keeps is the thing those six callers actually wanted: a bordered field with the unit inside it. If something later needs buttons, it should reach for `StepperTile` rather than reviving this.

PR #96, opened 2026-08-07
feat(backend): bucket day-scoped data in the user's timezone, server-side

Cawlumm:
**Stacked on #83.** Base is `pr83-timezone-base`, a copy of that PR's head pushed
here so this diff shows only the new work — #83 comes from a fork, so its branch
can't be a base directly. Once #83 lands, this retargets to `main`.

Merge this into #83 when it looks right, or land it after.

## Why

#83 makes the **client** compute local-day boundaries and send them as a range.
This adds the server-side half that comparable apps actually use, so a client can
send a bare `date=` and still get its own local day.

[Fitbit](https://community.fitbit.com/t5/Web-API-Development/Clarification-on-time-zones-and-date-time-values/td-p/925255)
stores the zone on the profile and renders every date in it with no zone attached.
[Oura](https://partnersupport.ouraring.com/hc/en-us/articles/29160913203219-Understanding-the-Different-Types-of-Oura-Days-in-Oura-API-Data)
defines Calendar Day as midnight-to-midnight in the zone the user's device reports.
[wger](https://github.com/wger-project/wger) uses Django's per-request active
timezone. None of them make each client do the date math — partly because every
client then has to implement it identically, which is already visible in #83 as
`localDayRange` existing in both `packages/shared` and `web/src/utils`.

The practical payoff: **already-installed mobile builds get correct bucketing with
no release**, because they already send `date=` and the server now reinterprets it.

## What changed

`user_settings.timezone`, an IANA name defaulting to `'UTC'` — exactly the
bucketing every existing row was written under, so the column is inert until a
client reports something real. Storing the **name** rather than an offset is what
makes DST self-correcting: `America/New_York` encodes both EST and EDT and resolves
per queried date.

Three endpoints resolve `date=` through it: food list, daily stats, and
`/food/history`.

**That last one is the reason this shouldn't wait.** History still grouped by
`substr(logged_at, 1, 10)` — the UTC date — while #83 moved the totals above it to
local ranges. Same screen, same user, two definitions of "day". Immich shipped that
exact split (server bucketing by `fileCreatedAt`, clients grouping by
`localDateTime`) and got a [date/time megathread](https://github.com/immich-app/immich/discussions/12650)
out of it. Grouping now happens per row in Go, in the user's zone — SQLite has no
IANA support, and a fixed hour offset is wrong twice a year.

Precedence is unchanged for existing callers: explicit `from`/`to` still wins, then
the user's zone, then the legacy path.

## Three things that would have failed silently

**The SQLite driver serializes a `time.Time`'s zone into the bound value.** A
boundary carrying `-0400 EDT` compares as text against UTC-stored rows and matches
nothing — the first run of these tests returned zero rows for exactly this.
Boundaries are normalized to UTC before they reach SQL.

**`golang:1.26-alpine` ships no `/usr/share/zoneinfo`.** Without
`import _ "time/tzdata"` every `LoadLocation` fails and falls back to UTC — the bug
this feature exists to fix, in a form that reproduces on no developer machine.

**Hermes doesn't expose the device zone to JavaScript.** A polyfilled `Intl`
reports `"UTC"` on every phone, so using it would have looked like the feature
worked while doing nothing. Mobile asks the OS via `expo-localization` (new
dependency, agreed); web uses `Intl`. Detection is injected, so compare-and-PATCH
lives once in `packages/shared`.

Zone names are validated by loading them, not by a pattern — the tz database is the
only authority on what exists, and it changes between releases. An unloadable name
is refused at write time rather than quietly making every later query fall back.

## Client behaviour

Silent sync on the settings fetch both apps already do at launch, so travel and a
moved device are picked up with no explicit step. Detection returning `null` sends
nothing and leaves the server's value alone.

**Known trade-off:** changing zones re-buckets history near midnight. Noon-anchored
food and weight are largely immune; real-instant workouts less so. This is the same
trade-off Fitbit and Oura accept. The permanent fix for diary data is a
`logged_on DATE` column — the MyFitnessPal/Cronometer model — which deserves its
own issue rather than growing this branch.

## Verification

- backend `go test ./...` all packages green
- web `tsc` + 122 unit tests; mobile `tsc` + 47; shared `tsc` + 61
- `time/tzdata` confirmed present in the binary's dependency graph

New tests cover what the old bucketing got wrong:

- a bare `date=` in a west zone (entry at 21:00 local lands on the right day)
- **UTC+14**, where noon-anchoring alone fails
- a **23-hour DST spring-forward day** — the boundary must be the next local
  midnight, not +24h
- history agreeing with the daily totals
- an unknown zone rejected with 400


Cawlumm:
Added `logged_on` in a second commit, per discussion — this is the deeper fix the first commit was approximating.

**The shape now:** day-scoped data (food, weight) stores the calendar day it belongs to; instants (workouts) keep a timestamp. The day is resolved once at write time — client-supplied wins, otherwise derived from `logged_at` in the user's zone — rather than recovered on every read.

Reads collapse to a column comparison:

| | before | after |
|---|---|---|
| food `ListByDay` | `substr(logged_at,1,10) = ?` | `logged_on = ?` |
| food `DailyMacros` | `substr(logged_at,1,10) = ?` | `logged_on = ?` |
| food `History` | `GROUP BY substr(...)` | `GROUP BY logged_on` |
| weight `UpsertForDay` | `logged_at >= ? AND < ?` | `logged_on = ?` |
| weight de-duplicate | `logged_at >= ? AND < ?` | `logged_on = ?` |

The two weight rows are an unreported bug fixed in passing: **"one entry per day" was a UTC timestamp window**, so it has always meant "per UTC day". Noon-anchoring hid it. `dayBounds` now has no callers and is deleted.

`History` also stops bucketing in Go — it's a plain `GROUP BY` again, on the same column the daily totals filter by, so the chart and the numbers above it *cannot* disagree.

**Two things worth a reviewer's eye:**

Backfill uses the UTC date prefix of `logged_at`. That's correct because every client writes these anchored at local noon, so the prefix already is the intended day for offsets within ±12h. UTC+13/+14 and third-party clients sending real instants can land a day out. It deliberately does *not* consult `user_settings.timezone` — the backfill runs at boot, before any client has reported a zone, so every user would read as UTC and the wrong day would be baked in permanently.

`seed/demo_data.go` inserts these tables directly rather than through the stores. Without setting the day there, the demo account's diary would have come up empty.

Tests moved with the semantics — bucketing is decided on write now, so they exercise the write path instead of inserting pre-stamped rows.

All suites green: backend `./...`, web (122), mobile (47), shared (61).

Cawlumm:
Cut back to one mechanism, per discussion — this reverses the `logged_on` direction rather than extending it.

**The premise was wrong.** `dayToIsoNoon` predates all of this work: food and weight were already written at local noon, so `substr(logged_at, 1, 10)` already yielded the day the user picked for any offset within ±12h — immune to travel and DST, consulting no timezone. And the diary-vs-chart mismatch I kept citing **did not exist on `main`**; list, stats and history all used the same `substr` expression. #83's range queries introduced that split, and `logged_on` was a third mechanism layered on to fix it.

So the column is gone — with the backfill, the re-derivation, the request field, and the seed/fixture columns.

## Reads — one rule

Every day query resolves the user's local day to an instant range via `localDayRange`. `substr` bucketing is gone from the codebase, and with it `WorkoutStore.CountOn` and weight's `dayBounds` windows plus the `-12h/+36h` widening that existed to guess an offset the server can now look up. Daily stats takes macros *and* the workout count from the same range, so the two halves of that payload can't describe different days.

## Writes — one method

`dayToInstant(day, previousIso?)` replaces three patterns:

| was | sites |
|---|---|
| `dayToIsoNoon(date)` — a noon anchor faking a day | 11 |
| `new Date(formData.date)` — UTC midnight | **1, and it was a live bug** |
| `applyDateKeepTime(...)` — correct, but local to one screen | 1 |

**The bug:** `mobile/app/(tabs)/workouts/new.tsx:176` still had the UTC-midnight parse that #83 fixed on web, dating workouts a day early for anyone west of UTC. Fixed, with a regression test.

Passing the previous timestamp moves the day and keeps the time-of-day. Without it the default is local noon — kept deliberately, now stated as a default rather than an encoding trick.

Three sites deliberately stay off it: two live session starts and `AddWorkoutModal` use a real instant because they're happening now; the two web workout edits preserve the original because **neither screen has a date field** (checked). Different question, not the same one answered differently.

`dayToIsoNoon`, `applyDateKeepTime` and a local `startedAtToDate` (which reimplemented `isoToDayInput`) are deleted, so there's no second method left to reach for.

## Kept

`user_settings.timezone`, the tzdata embed, zone validation and client detection. That's what makes **UTC+13/+14** correct — the one case a single anchor genuinely cannot cover, and the reason the zone is worth storing at all.

## Verification

backend `go build`/`go vet`/`go test ./...` green · web tsc + **125 tests** · mobile tsc + 47 + lint 0 errors · shared tsc + **59 tests**

(`npm run lint` in web fails on this branch only because the ESLint setup is #95 on `main` and this stack predates it.)

Cawlumm:
Superseded by #83 — closing.

Everything here has been pushed onto `fix/timezone-app-wide`, so #83 now contains the complete change: the original timezone fixes plus the server-side bucketing, the single `dayToInstant`/`instantToDay` pair, and the mobile UTC-midnight bug fix.

It went up as a **fast-forward** (`e427d2d..5649abb`) — nothing rewritten, and @DisplacedForest's 9 commits keep their authorship. Merging #83 rather than this PR means the contribution is credited to them, which is the point.

Review and testing should happen on #83.

PR #102, opened 2026-08-10
Make web an npm workspace member so it can import @lyftr/shared

Cawlumm:
web has carried a hand-maintained copy of everything in `packages/shared`. The blocker was never the code — it was the plumbing. This PR fixes the plumbing and migrates `programUtils` as a canary. **No user-visible behavior changes.**

`programUtils` is the right canary because it is a *runtime* import. Types erase at compile time and would prove nothing about what actually ends up in the bundle.

## What moved

| | Before | After |
|---|---|---|
| Workspaces | `packages/*`, `mobile` | + `web` |
| Lockfiles | `package-lock.json` + `web/package-lock.json` | one root lockfile |
| web Docker context | `./web` | repo root |
| `.dockerignore` | `web/.dockerignore` only | new root file; `web/.dockerignore` deleted |
| CI `web` paths filter | `web/**` | + `packages/shared/**`, `package.json`, `package-lock.json` |

## Three things worth reviewing closely

**1. The dedupe block lives in one shared module.** `web/vite.resolve.ts` is imported by *both* `vite.config.ts` and `vitest.config.ts`. Vitest ignores `vite.config.ts` entirely when a `vitest.config.ts` exists, so a resolve block written into only one of them yields a green build with a red test run — and it surfaces as `Invalid hook call`, which names nothing useful.

**2. The root `.dockerignore` is load-bearing, not tidiness.** BuildKit reads `.dockerignore` from the *context root*, so `web/.dockerignore` stopped applying the moment the context moved up. Without a root one, `COPY web/ ./web/` drops a stale host `node_modules` (React 18, host-platform binaries) on top of the one `npm ci` just built — and the image builds **green against the wrong dependency tree**. Verified inside the builder stage:

```
expo: ABSENT (good)
./node_modules/react/package.json        <- exactly one copy
/app/node_modules/react/package.json 19.1.0
node_modules/@lyftr/shared -> ../../packages/shared
```

`npm ci -w lyfter-web --include-workspace-root` keeps Expo's toolchain out of the image. `mobile/package.json` is copied only because `npm ci` validates the lockfile against every declared workspace — its dependencies are never installed.

**3. The CI paths-filter edit is the successor to the fork guard.** Today a `packages/shared/**`-only PR runs neither web's unit suite nor e2e. Once web imports shared, that gap means a shared change could break the web app with fully green CI. It lands in this PR rather than a later one for that reason.

## Dependency churn — please skim

Collapsing two lockfiles into one re-floats every caret range in web. Nine direct deps moved, all within their declared ranges:

| Package | Old | New |
|---|---|---|
| `lucide-react` | 1.8.0 | **1.31.0** |
| `@playwright/test` | 1.59.1 | 1.62.1 |
| `axios` | 1.15.0 | 1.18.1 (now dedupes with the shared client) |
| `@tanstack/react-virtual` | 3.13.24 | 3.14.9 |
| `react-zxing` | 2.1.0 | 2.2.0 |
| `react-router-dom` | 6.30.3 | 6.30.4 |
| `autoprefixer` | 10.4.16 | 10.5.4 |
| `eslint-plugin-react-refresh` | 0.5.3 | 0.5.4 |
| `postcss` | 8.5.10 | 8.4.49 *(sideways — another workspace member pins it; still satisfies `^8.4.32`)* |

`lucide-react` is the one that touches every screen. `tsc` proves every imported icon *name* still exists; it does not prove they still *render* — an icon resolving to an empty component would type-check and build clean. So icons were counted across seven routes: **0 empty, 0 console errors.**

## Verification

| Check | Result |
|---|---|
| `npm run type-check` / `lint` | clean, warnings 79 → **69** |
| `npm run test:unit` | 131 passed |
| `npm run build` | clean |
| `npm run test:e2e` | **88 passed**, 1 skipped (chromium + mobile) |
| `npm run shared:type-check` / `shared:test` | 113 passed |
| mobile `type-check` / `test` | 47 passed |
| `go test ./controllers/ ./db/ ./routes/` | pass |
| `docker compose build frontend` | pass |
| `docker build -f fly/Dockerfile .` | pass |

The fly build matters because it is **production and is skipped on PRs** — it deploys on every push to `main`, so it was built locally. All three builds emit the same bundle (`index-Can3woFf.js`).

## Found along the way — not fixed here

`backend/build.sh` has CRLF endings in the working tree (`core.autocrlf=true`, no `.gitattributes`), so **`docker build -f fly/Dockerfile .` fails on a Windows checkout** with `set: line 6: illegal option -`. Pre-existing and unrelated to this diff — git stores the file as LF, so CI and `flyctl` on Linux are unaffected; only local Windows production builds break. It was masked until now by a cached layer. A `.gitattributes` with `*.sh text eol=lf` would fix it; left out of this PR to keep the diff to one concern.


Cawlumm:

PR #103, opened 2026-08-10
Take web's domain types from @lyftr/shared

Cawlumm:
Stacked on #102 — **merge that first.**

`web/src/types.ts` and `packages/shared/src/types.ts` differed by exactly two comment lines. Deleting web's copy removes the last place an API shape can be described twice and disagree.

## The one judgement call

The two apps had settled on different import styles:

```ts
// mobile — domain types by name, alongside functions
import { weightShort, type Exercise } from '@lyftr/shared'

// web — namespaced, at 28 call sites
import * as types from '../types'
//   ...types.Workout, types.Program, types.ActiveSession
```

Making web adopt mobile's spelling means renaming every `types.Foo` reference — roughly 500 lines of churn that would bury the actual change and make the diff unreviewable. So shared's index also re-exports the module as a namespace:

```ts
export * from './types'
export * as types from './types'
```

Both spellings resolve to the same declarations. Each app keeps its convention and the diff is one line per file.

## Verification

Types erase at compile time, so this should produce the same program — and it does. The build emits a **byte-identical bundle** (`index-Can3woFf.js`) before and after, which is a stronger guarantee than the test suite gives.

| Check | Result |
|---|---|
| type-check (web / shared / mobile) | clean |
| web unit | 131 passed |
| shared | 113 passed |
| lint | clean, 69 warnings |
| build | identical output hash |
| e2e | 88 passed, 1 skipped |


Cawlumm:

PR #104, opened 2026-08-10
Move the pure utils into @lyftr/shared and delete the forks

Cawlumm:
Stacked on #103 (which is stacked on #102) — **merge in order: 102 → 103 → this.**

Every util that was pure logic now has exactly one definition.

## What moved

| Util | Was | Cost of the fork |
|---|---|---|
| `dateUtils` | web ↔ shared | byte-identical, kept in step by a bespoke byte-comparison test |
| `number` | web ↔ mobile | two comment lines apart |
| `workoutSets` | web ↔ mobile | two comment lines apart |
| `MACRO_COLORS` | `web/utils/macroColors.ts` ↔ `mobile/.../nutritionMeta.ts` | identical |
| `EQUIPMENT_LABEL` | inside both `exerciseUtils` | identical |

## `exerciseUtils` splits rather than moves

The *lookup rule* is identical on both sides, so `resolveMuscleSlugs(name, table)` is shared. The **tables are not**, and this is deliberate:

- **Slugs** — web draws with `react-body-highlighter`, mobile with `react-native-body-highlighter`, and they disagree on vocabulary. The web library splits front/back deltoids and has `abductors`; the RN one has a single `deltoids` and neither `abductors` nor `middle-back`. One shared table would emit slugs a renderer silently ignores — a diagram that quietly highlights nothing.
- **Colour** — Tailwind class strings on web, a `{chip, border, text}` tint on mobile, because RN needs the text colour as an inline style to beat stylesheet ordering. Same palette, different delivery.

Each side gained a test for exactly the part that is its own (web's border-stripping and front/back deltoid split; mobile's `abductors → gluteal` remap).

## The fork guard is gone, and nothing directly replaces it

`dateUtils.fork.test.ts` existed because two copies had to be compared. With one copy there is nothing to compare. Its successor already landed in #102: `ci.yml`'s `web` paths filter includes `packages/shared/**`, so editing a shared util now runs web's unit suite, build and e2e alongside mobile's.

`CLAUDE.md` told contributors to hand-copy `dateUtils` between the two files. That instruction is now wrong, so it's rewritten in this PR rather than left to rot.

## Dead code removed

- **`web/src/utils/numberUtils.ts`** — `isPositiveNumber` is referenced by nothing except its own test. That's why the test kept passing and nobody noticed.
- **`weightLabel`** in shared — never imported by either app.
- **`web/tsconfig.build.json`** — it existed only because the fork guard used `node:fs`/`__dirname` and `@types/node` is absent in the image. Nothing in `web/src` touches a node builtin now, so `build` goes back to plain `tsconfig.json`. Verified by actually building the image, not by reading the comment.

## Test counts move rather than shrink

| Suite | Before | After |
|---|---|---|
| web | 131 | 111 |
| mobile | 47 | 33 |
| shared | 113 | **131** |

Net −16. Those were duplicated assertions being run twice against two copies of the same function.

## Verification

type-check (web / shared / mobile) · all three unit suites · lint (69 warnings, unchanged) · build · **88 e2e** · Go tests · `docker compose build frontend` (which is what confirms the `tsconfig.build.json` removal).


Cawlumm:

PR #105, opened 2026-08-11
Consolidate the duplicated page helpers, hooks and design tokens

Cawlumm:
**PR A of 3.** Stacked on #104 → #103 → #102. Merged bottom-up, the whole stack lands on `main` as one commit.

Mechanical de-duplication, no behavior changes.

## Helpers that existed more than once

Two of them were duplicated **inside** each app as well as across them — four copies of the same three lines:

| Helper | Copies | Where |
|---|---|---|
| `restLabel` | **4** | web ProgramDetail + WorkoutDetail, mobile programs/[id] + workouts/[id] |
| `formatElapsed` | **4** | web Layout + ActiveWorkout, mobile SessionPill + workouts/active |
| `entryToResult` / `savedToResult` | 2 | LogFood ↔ nutrition/log |
| `muscleHex`, `muscleRoast` | 2 | both dashboards — including the roast copy strings |
| `upNext` | 2 | → `nextStartableDay(programs)` |
| `wLabel` | 2 | → `targetWeightLabel(lbs, unit)` |
| `range` | 2 | → `numericRange(values)` |
| `useCountdown`, `useNumericText` | 2 | identical bodies |
| `formatVersion` | web only | mobile printed the raw `git describe` string — now formats the same |

`entryToResult` is the one worth reading closely: it divides a logged entry back down to a single serving, because the detail view multiplies by `servings` again. It lived on one screen per platform with **no link between them**, so a change to either side would have silently mis-logged food on the other.

Moving the two hooks makes `packages/shared` import `react` for the first time — already allowed by its `peerDependencies`, and safe now that both apps are on 19.1.0.

## Design tokens: three copies → one

`web/src/index.css` declared the CSS custom properties, `web/tailwind.config.ts` repeated the brand hexes, `mobile/src/theme/theme.ts` repeated all of them again.

```
packages/shared/src/theme/tokens.ts   ← the only copy
        ├── web/tailwind.config.ts    → palette + addBase() emits :root / .dark
        └── mobile/src/theme/theme.ts → imports the same object
```

`index.css` now declares no values at all. The config **deep-imports** `tokens.ts` rather than the package index, so loading Tailwind doesn't drag axios and the API client into PostCSS — `tokens.ts` is deliberately dependency-free.

I checked this one twice, because a wrong token is a sitewide visual regression:

**1. The built CSS emits the same nine variables with the same values.**

```css
:root{--surface-base: #f8fafc;--surface-raised: #ffffff; … color-scheme:light}
.dark{--surface-base: #070d1a;--surface-raised: #0d1629; … color-scheme:dark}
```

**2. A browser resolves them.** All nine computed on `document.documentElement`, in both themes, plus `color-scheme` and the painted `body` background:

```
light  color-scheme: light  bodyBg: rgb(248, 250, 252)  mismatches: []
dark   color-scheme: dark   bodyBg: rgb(7, 13, 26)      mismatches: []
```

## Two bugs caught while doing this

Both from checking rather than assuming, and both would have shipped silently:

- **`formatElapsed` zero-pads its minute** (`05:30`). I'd written the shared version unpadded — the digits would have jumped as the timer crossed 9:59.
- **`MUSCLE_ROAST` has a `'full body'` entry** that a truncated diff had hidden, so my first shared copy was missing a line of copy.

## Verification

type-check (web / shared / mobile) · unit suites **shared 155, web 108, mobile 33** · lint clean on both apps (0 errors) · build · **88 e2e** · Go tests · theme verified in-browser in both modes.

The two hook tests stay in `web/src/hooks/` — exercising a React hook needs jsdom + `@testing-library`, which shared's plain-node jest setup doesn't carry. `ci.yml` runs web's unit job on any `packages/shared/**` change, so a shared edit still trips them.

## Not in this PR

`calcVolume` is still duplicated and still disagrees with the backend, which filters `is_warmup = 0`. Mobile's copy also coerces with `Number(x) || 0` to avoid a `NaN` that hard-crashes `react-native-svg`; web's doesn't. Both go in **PR B** with the API client and stores, since the fix changes displayed numbers.


Cawlumm:

PR #106, opened 2026-08-11
Run web on the shared API client and stores

Cawlumm:
**PR B of 3.** Stacked on #105 → #104 → #103 → #102. Merged bottom-up, the stack lands on `main` as one commit.

Web's API client and three of its four stores were reimplementations of code `packages/shared` already had and mobile already used. They're now the same code, bound to a web `localStorage` adapter instead of mobile's Keychain/AsyncStorage one.

```
web/src/lib/storage.ts   StorageAdapter over localStorage
web/src/lib/lyftr.ts     client + server/settings/workoutSession stores
                         (mirrors mobile/src/lib/lyftr.ts)
```

`services/api.ts` and the three store modules become **re-export shims**, so ~29 API call sites and ~40 store consumers are untouched — the diff shows what moved rather than a rename of everything referencing it. Collapse later.

**Net: −1107 lines, +362.**

## The auth store is deliberately not here

Web reads `localStorage` during `create()`, so the first render already knows you're signed in. The shared factory starts unauthenticated and hydrates. `App.tsx`'s catch-all is `<Navigate to="/login">`, which performs a history **replace** — pre-hydration that destroys a deep link before it resolves. That gets its own PR and its own test.

## Behaviour that genuinely changes

| Before | After |
|---|---|
| **No request timeout at all** — a dropped or blocked request hung the UI forever | 20s global, 60s for the ~820 KB exercise list |
| `sync()` (exercise re-seed) waited indefinitely | explicit **120s** — 20s would report failure for work the server finishes |
| One hardcoded CORS string on any network failure | `networkFailureMessage` distinguishes blocked cleartext / untrusted cert / unreachable |

Verified live rather than assumed:

```json
{ "globalTimeout": 20000, "syncTimeout": 120000 }
```

## Upgrade safety

`STORAGE_KEYS` was written to match web's existing key names, so tokens, server URL and an in-progress workout all survive. Checked two ways — a browser probe confirms `access_token` / `refresh_token` / `user` land under the historical keys, and a unit test asserts `WORKOUT_SESSION_KEY` is still `'lyftr_active_session'`. If those had drifted, everyone mid-workout at deploy time would lose their session.

## Hydration — and the bug it prevents

`main.tsx` now awaits persisted state before the first render. This matters more than a flicker: `ActiveWorkout`'s gym-layout election is a **mount-only** effect (`[]` deps), so a `workout_layout` still reading `'list'` when it fires leaves a gym-mode user in the list layout **and never re-runs**.

That's why `createSettingsStore` gained `hydratePrefs()` — the three device-only prefs, no network, before paint. A rejection still renders: usable defaults beat a blank page if `localStorage` is unavailable (Safari private mode).

## `calcVolume` — the one number that changes

Moved to shared with two fixes folded in:

1. **Warm-up sets excluded**, matching `backend/stores/workout.go`, which computes `SUM(s.reps * s.weight) ... AND s.is_warmup = 0`. Six client call sites summed unfiltered. Nothing writes `is_warmup` yet, so every row is `0` and the two agree **today** — which is exactly why this is the safe moment to align them.
2. **reps/weight coerced** — a guard mobile had and web didn't. One `NaN` poisons the whole sum; web drops the geometry silently, native turns it into a hard `react-native-svg` crash.

## Also added to shared

`exerciseAPI.seedStatus()` and `sync()`, plus `isMixedContentBlocked` / `MIXED_CONTENT_WARNING` moved in from web's server store. The guard now tests `window.location` rather than `window` — under Hermes a `window` global can exist without a `location`, so checking only `window` would throw on the property read instead of returning false. Always false on native.

## Tests follow the code

Pure reducers and weight math live in shared. Web keeps only what needs a browser: the localStorage binding, and `isMixedContentBlocked`, which reads `window.location.protocol` and is unreachable in shared's node jest.

| Suite | Count |
|---|---|
| shared | 161 |
| web | 73 |
| mobile | 33 |

## Verification

type-check (web / shared / mobile) · lint clean both apps, budget **69 → 55** · build · **88 e2e**, including gym-layout persistence and session-restore-after-refresh · Go tests · live probe of login, storage keys, `seedStatus` (873 exercises) and both timeouts.


Cawlumm:

PR #107, opened 2026-08-11
Share the day-cycle reducers and the elapsed-workout timer

Cawlumm:
Stacked on #106. Two things a re-audit caught that PR A missed.

**I listed the `ProgramDaysEditor` reducers as "movable now" in the review and then didn't move them, without flagging it.** This closes that gap rather than leaving it implied.

## Day-cycle reducers

Both apps' `ProgramDaysEditor` had its own `reindex` / `addDay` / `removeDay` / `moveDay` / `updateDay`, identical after comments. They move to `packages/shared/src/utils/programDrafts.ts` as pure array operations. The `expanded` row bookkeeping stays in each component — that's UI state, not data.

They're **generic over the day type** rather than importing a concrete draft. The two editors' *set* drafts still disagree on field names:

| | web | mobile |
|---|---|---|
| `DaySetDraft` | `target_reps` / `target_weight` (matches backend) | `reps` / `weight`, mapped at submit |

None of these five operations touch sets, so the difference is irrelevant here. Sharing generically avoids a rename of mobile's program create/edit path that **this change cannot verify end-to-end** — mobile builds are EAS-only, so there's no Maestro run available from here. The divergence is real and worth fixing, separately, where it can be tested properly.

One behavioural detail worth reviewing: `moveDayBy` returns the **same array reference** when the move would fall off either end, which is how the caller knows to skip its expanded-row bookkeeping. The previous code returned early instead. There's a test asserting *identity*, not equality — a structural clone would pass the obvious version of that assertion.

## Elapsed timer

Four copies of the same effect — `Layout`, `ActiveWorkout`, `SessionPill`, `workouts/active` — became `useElapsedSeconds(startedAt)`. It recomputes from the absolute start each tick, so it neither drifts nor needs catching up after a backgrounded tab or a locked phone.

**One real fix folded in:** three of the four keyed the effect on `session`, so the interval was torn down and rebuilt every time *any* part of the session changed — completing a set, editing a weight. The hook keys on the timestamp instead. Same displayed value, without the churn.

## Verification

type-check (web / shared / mobile) · **shared 171, web 73, mobile 33** — both apps' `ProgramDaysEditor` suites (8 tests each) now exercise the shared reducers · lint clean, budget **55 → 54** · build · **88 e2e** · Go tests.

The timer was checked live rather than by inspection:

```json
{ "clocksAtStart": ["01:06"], "clocksAfter2_6s": ["01:09"], "ticked": true, "pageErrors": [] }
```

A session seeded 65s in the past reads `01:06` and advances to `01:09`, so the interval genuinely runs. That check exists because a broken dependency array would render `00:00` forever and **every existing e2e would still pass** — they only assert the pill exists.


Cawlumm:

PR #108, opened 2026-08-11
Move all shared logic into @lyftr/shared

Cawlumm:
Supersedes #102, #103, #104, #105, #106, #107 — all six are folded in here and closed. Seven commits, reviewed and tested separately, landing as one merge.

**129 files (excluding lockfiles), +1,766 / −2,797.** 20 files added, 20 deleted.

Web and mobile each carried their own copy of the domain logic. Now there is one copy, and what stays local stays local for a reason.

## The seven commits

| # | Commit | What |
|---|---|---|
| 1 | Make web an npm workspace member | Workspace, Docker context, CI plumbing |
| 2 | Take web's domain types from shared | `types.ts` deleted — byte-identical bundle |
| 3 | Move the pure utils into shared | dateUtils, number, workoutSets, colours + dead code |
| 4 | Consolidate page helpers, hooks, tokens | 4×`restLabel`, 4×`formatElapsed`, food math, design tokens |
| 5 | Run web on the shared client and stores | API client + 3 stores, −1,107 lines |
| 6 | Share day reducers + elapsed timer | Program editors, 4× elapsed effect |
| 7 | Fix the ten review defects | See below |

## The rule that decides what moves

Not "is it useful to both" — that always answers yes and makes shared a dumping ground. The test is **what does it import**:

| Imports | Home |
|---|---|
| nothing platform-specific | shared |
| `window`, `document`, DOM, `import.meta` | web |
| React Native, Expo, Hermes | mobile |
| needs a platform thing, logic is general | shared, platform thing **injected** |

That last row is the `StorageAdapter` / `TimezoneDetector` seam. It's why the API client and every store could move.

**Deliberately still local:** `exerciseUtils` colour + slug tables (the two body-highlighter libraries disagree on their slug vocabulary — one shared table would emit slugs a renderer silently ignores), `useBodyScrollLock`/`useEscapeKey` (DOM), `useStableCallback` and `workoutOutcome` (mobile-only concepts), each platform's storage adapter.

A boundary audit confirms the end state: **nothing either app redefines that shared already exports**, and every remaining local file touches a platform API or a platform library's vocabulary.

## Bugs found and fixed along the way

The consolidation was the point; these were the payoff.

- **`exerciseAPI.list` cached rejections forever.** The in-flight promise was cleared only on success, so one failed bulk fetch broke the exercise picker until a page reload — and `clearCache()` has no callers.
- **`useTheme` held per-component state.** Four components each owned a copy; toggling in Settings re-themed the page but left Layout's menu label stale.
- **`calcVolume` disagreed with the backend**, which filters `is_warmup = 0`. Now aligned, along with the set count and per-exercise rows so the card can't contradict itself.
- **`greeting()` read a module-load date** — a tab open across midnight kept yesterday's header, week boundary and heatmap "today".
- **Web had no request timeout at all.** Now 20s, with a 60s bulk bound for reads over 50 rows and 120s for the exercise re-seed.
- **Only half the food serving round-trip was shared**, leaving the two directions free to drift.
- Mobile fixes web never got are still queued (`useRestTimer`'s one-frame banner unmount, `useServerInfiniteList`'s missing request-race guard) — they need the auth store first.

Two of the ten were **introduced by this work** and caught by adversarial review: `reset()` wiping hydrated device prefs, and the volume/set-count inconsistency. Both have tests that were confirmed to fail against the buggy version before the fix went back in.

## Upgrade safety

Storage keys are unchanged, so tokens, server URL and an in-progress workout all survive. Verified two ways — a browser probe (tokens under the historical keys) and a test pinning `WORKOUT_SESSION_KEY` to `'lyftr_active_session'`. Drift there would sign everyone out and drop live workouts at deploy time.

## Verification

| Check | Result |
|---|---|
| type-check (web / shared / mobile) | clean |
| unit tests | shared **194**, web 73, mobile 33 |
| lint | 0 errors both apps; web budget 79 → 54 |
| build | clean |
| e2e | **88 passed**, 1 skipped |
| Go tests | pass |
| `docker compose build frontend` | pass |
| `docker build -f fly/Dockerfile .` | pass (production, skipped by PR CI) |

Beyond the suites, the things a green test run would not have caught were checked in a browser: theme tokens resolve in both modes, the elapsed timer actually ticks, lucide icons render on seven routes, the theme toggle propagates across components, and the client's timeouts are what they claim.

## Known gap

**Mobile has no end-to-end run.** Builds are EAS-only, so `_layout.tsx`, the program editor and the nutrition screens are covered by type-check and jest only. A Maestro pass before merge would close that.


.github/workflows/ci.yml
```diff
@@ -60,6 +60,13 @@ jobs:
               - '.github/workflows/ci.yml'
             web:
               - 'web/**'
+              # Web imports @lyftr/shared, so a shared-only edit must run web's unit
```
Cawlumm, line 63: So this will make sure that when changes to shared code happens, Ci will run for it.

fly/Dockerfile
```diff
@@ -9,12 +9,19 @@
...
+# web is an npm workspace member, so this installs from the root lockfile. `-w lyfter-web`
+# limits it to web's slice; mobile/package.json is copied only because `npm ci` validates
+# the lockfile against every declared workspace. Mirrors web/Dockerfile — keep them in step.
 FROM --platform=$BUILDPLATFORM node:20-alpine AS frontend
 WORKDIR /app
-COPY web/package.json web/package-lock.json ./
-RUN npm ci
-COPY web/ .
-RUN npm run build
+COPY package.json package-lock.json ./
```
Cawlumm, line 17: Make sure to test pulling docker images and using them once built either locally or push to dockerhub locally and pull locally to test.

fly/Dockerfile
```diff
@@ -37,7 +44,7 @@ FROM nginx:alpine
 RUN apk add --no-cache ca-certificates wget
 
 COPY --from=backend /build/lyftr-api     /app/lyftr-api
-COPY --from=frontend /app/dist           /usr/share/nginx/html
+COPY --from=frontend /app/web/dist       /usr/share/nginx/html
```
Cawlumm, line 47: What is the reason for this.

mobile/app/(tabs)/nutrition/log.tsx
Cawlumm: 'FoodLog' is defined but never used, 'FoodLog' is defined but never used

mobile/app/(tabs)/weight/index.tsx
Cawlumm: 'subDays' is defined but never used
CI / Mobile + shared checks
Check warning on line R4
Check warning:
'subDays' is defined but never used

.github/workflows/ci.yml
Cawlumm: Yes — that is exactly what it does, and it is the load-bearing part of this PR's CI change.

Before, the `web` filter was only `web/**`. A PR touching only `packages/shared/**` ran the mobile job and nothing else, so a shared change could break the web app with fully green CI. Now that web imports shared, that gap would be a real hazard.

This entry is also the replacement for the guard we deleted. `dateUtils.fork.test.ts` used to byte-compare web's copy of `dateUtils` against shared's; with one copy there is nothing left to compare, so CI coverage is what protects it instead. That is why it landed in the first commit rather than later.

`package.json` and `package-lock.json` are in the same filter for the same reason — a dependency change now affects web's install too.

fly/Dockerfile
Cawlumm: Done — built both image paths locally and actually ran them, not just built them.

**fly/Dockerfile** (the single-container production image, which PR CI never exercises since the deploy job is skipped):

```
docker build -f fly/Dockerfile -t lyftr-fly:pr108 .
docker run -d -p 8088:80 -e JWT_SECRET=... lyftr-fly:pr108

GET /                     -> 200   <title>Lyfter - Workout Tracking</title>
GET /assets/index-*.js    -> 200   (1,443,419 bytes)
GET /api/v1/info          -> 200   {"name":"lyftr","version":"dev"}
POST /api/v1/auth/login   -> 200   (demo user seeded inside the container)
/api/v1/me, /settings, /workouts, /exercises -> 200
```

So nginx serves the SPA and proxies to the Go backend on localhost:3000 inside the same container, and the DB seeds on first boot.

**docker compose** (the two-image stack) — same checks, plus a browser pass against the *production* bundle rather than the dev server:

```json
{ "dashboard": { "greeting": "Good morning", "hasSections": true },
  "deepLinkSurvived": true,
  "tokens": { "base": "#070d1a", "txPrimary": "#f1f5f9" },
  "pageErrors": [] }
```

That last one matters for this PR specifically: `--surface-base` resolving in the built CSS confirms the Tailwind `addBase` plugin is emitting the shared design tokens into the shipped stylesheet, and `deepLinkSurvived` confirms the hydration gate works in the minified bundle.

One thing worth recording: on my first attempt I mapped compose to port 8089 and the browser login returned **403**. That was my harness, not the image — compose sets `CORS_ORIGIN=http://localhost`, so an origin of `http://localhost:8089` is correctly rejected. curl passed because curl sends no `Origin` header. Re-running on the default port worked. Worth knowing if you ever run the stack on a non-80 port locally: set `CORS_ORIGIN` to match.

I did not push to Docker Hub — these were built and pulled from the local daemon. Say the word if you want a registry round-trip too.

fly/Dockerfile
Cawlumm: This is the `COPY --from=frontend /app/web/dist` line, and the reason is the workspace move.

Before this PR, `fly/Dockerfile`'s frontend stage did `COPY web/package.json ...` then `COPY web/ .` — so web's source landed directly at `/app`, and Vite emitted to `/app/dist`.

Now web is an npm workspace member, so the install has to happen at the workspace root to resolve `@lyftr/shared`. The stage copies the root manifests to `/app`, web's source to `/app/web/`, and runs `npm run build -w lyfter-web`. Vite runs with its root at `/app/web`, so the output is `/app/web/dist`.

The runtime stage just has to follow the build output — hence `/app/dist` → `/app/web/dist`. If this line had been missed the image would still build fine and then serve an empty nginx root, which is the sort of thing that only shows up at deploy. It is covered by the container run in the thread above: `GET /` returns the real `index.html` and the JS asset is 1.44 MB, so the assets are genuinely in place.

Same reasoning applies to `web/Dockerfile`, which has the matching `COPY --from=builder /app/web/dist`.

mobile/app/(tabs)/nutrition/log.tsx
Cawlumm: Fixed in `1badebd`. `FoodLog` became unused when `entryToResult` moved to `@lyftr/shared` — the type went with it and I left the import behind.

I treated this and the `subDays` one as a class rather than two one-offs, and swept all three packages for the same thing:

| | unused vars |
|---|---|
| mobile | 3 — `FoodLog`, plus `useEffect`/`useState` in `SessionPill.tsx` |
| web | none |
| shared | 1 — a `hours` helper in `dateUtils.test.ts` |

`SessionPill` is the same cause as this one: its elapsed-timer effect became `useElapsedSeconds`, so the React hooks it imported were orphaned. Web was already clean because I removed its orphans when the Dashboard helpers moved.

Mobile lint warnings are 16 → 12. The remaining 12 all predate this PR and I left them alone: `require()`-style imports and `import/first` in three test files, and four `react-hooks/exhaustive-deps` in `RestTimerBanner` and `WorkoutSessionLayer`. Happy to take those in a separate pass if you want them gone.

mobile/app/(tabs)/weight/index.tsx
Cawlumm: Fixed in `1badebd`, with one correction worth stating: **this one is not mine.** `subDays` is unused on `origin/main` too —

```
git show origin/main:'mobile/app/(tabs)/weight/index.tsx' | sed -n 4p
import { format, subDays } from 'date-fns'
```

It only surfaced on this PR because the `web`/`mobile` paths filter now runs the mobile job for a shared change, so CI is annotating a file this branch happens to touch. I removed it anyway — it is dead either way and the same one-line kind of cleanup as the rest of this pass — but I did not want to imply the refactor caused it.

PR #110, opened 2026-08-14
feat: let a self-hoster close registration, and stop shipping a public account

### LINKED ISSUE #86: Lock registration after the owner signs up (self-host: closed / first-user-only registration)
## What problem does this solve?

`POST /api/v1/auth/register` (`backend/routes/routes.go:29` → `backend/controllers/auth.go:14`) is public and unconditional. Any Lyftr instance reachable from the internet — which is the normal setup once someone puts it behind a reverse proxy — will happily accept unlimited signups from anyone who finds it. A self-hoster running a single-user instance has no way to say "I'm the only account here, stop taking new ones."

Today the only workarounds are outside the app: a proxy rule blocking the register path, or basic auth in front of everything. Both are easy to get wrong and neither is discoverable.

## Describe the feature

A server-side switch for whether registration is open, exposed as config:

```
REGISTRATION=open          # today's behaviour, default
REGISTRATION=closed        # /auth/register always 403
REGISTRATION=first-user    # open only while the users table is empty
```

`first-user` is the mode worth having as the recommended one: it solves the bootstrap problem cleanly. With a plain on/off flag you either start open (and race a scraper to your own signup) or start closed (and can't create your own account at all).

Sketch:

- **Config** — one entry in `backend/config/config.go` next to `CORS_ORIGIN`, same `getEnv` pattern.
- **Enforcement** — `Register` returns 403 before doing any work when registration is closed. This is the part that actually matters; a scraper posts to the API directly and never sees the UI.
- **`first-user` needs a user count** — `UserStore` has no `Count()` today (`backend/stores/user.go`), so that's a small addition.
- **Advertise it on `/api/v1/info`** (`backend/controllers/server.go`) — that endpoint already exists as the public, unauthenticated probe both clients call during "test connection", so adding `registration_open: bool` costs nothing and needs no auth.
- **Clients hide the affordance** — the "Create account" link on `web/src/pages/Login.tsx` and `mobile/app/(auth)/login.tsx`, and a direct visit to `/register` should explain rather than fail on submit. UI hiding is politeness, not enforcement.

## Who benefits?

Anyone running a single-user or family instance on a public hostname — which is most self-hosters once they've followed the reverse-proxy docs. Related friction has come up before in #31 and #14.

## Anything else?

Things to settle when this is picked up:

- **Seeded users interact with `first-user`.** `backend/seed/users.go` seeds a demo account; if that runs before the owner registers, the table isn't empty and the owner is locked out. The count probably needs to ignore seeded/demo accounts, or `first-user` needs to be incompatible with demo seeding.
- **Changing the setting must not require a rebuild** — env var + container restart is fine, but worth being explicit in the docs.
- Should a closed instance return 403 or 404 on `/auth/register`? 403 with a clear message is friendlier to a legitimate second user; 404 leaks less to a scanner.
- Out of scope unless someone asks: invite codes, admin-managed user list, per-user disable. Those are a different, bigger feature.

Cawlumm:
Closes #86.

## What

`REGISTRATION=open|first-user|closed`, enforced in the API and advertised on `/api/v1/info` so both clients can hide the affordance. Plus the thing that makes it mean something: demo seeding is now behind `DEMO_MODE`.

## Why the second switch is in the same PR

Every install seeded `demo@lyftr.local` with a password published in the README. That makes closed registration theatre — an attacker does not need to register — and it makes `first-user` impossible, because the users table is never empty. `DEMO_MODE` defaults to `ENV == "development"`, so contributors keep their login and production installs get no default account at all.

Existing installs are **not** cleaned up automatically: the row may hold real workouts by now. The backend logs a warning at startup while it exists, pointing at Settings → Delete account.

## Decisions

| | |
|---|---|
| Default `REGISTRATION` | `open` — an upgrade must not lock anyone out, and the e2e suite registers a user per worker |
| Closed response | 403 with a message. `apiErrorMessage` surfaces `response.data.error` verbatim, so both clients show it with no client work. A 404 gets mapped by that same function to *"Server URL looks misconfigured"* — wrong and confusing for a legitimate second user |
| Invalid value | Startup failure, never a fallback to open. `REGISTRATION=frst-user` silently meaning "anyone may sign up" is the one failure mode that leaves someone believing they are locked down |

## Enforcement

- Rejected **before binding or hashing** — bcrypt on an unauthenticated public endpoint is a free CPU burn for whoever finds a closed instance.
- `first-user` re-counts **inside the insert transaction** (`stores.UserStore.CreateFirst`). The handler's pre-check alone loses the exact race the mode exists to prevent: two requests to a fresh instance can both observe an empty table and both be waved through.
- `registration_open` is optional on the client type. A backend older than this omits it, and absent reads as **open** — a new client must never hide a working signup link against a server the user cannot upgrade.

## Tests

`backend/controllers/auth_test.go` (all three modes, the race guard, and that a closed instance writes nothing), `config_test.go` (`getEnvBool` + the validation set), `server_test.go` extended for `registration_open` flipping under `first-user`, and a shared jest case pinning the absent-field rule.

Each guard was checked by mutation — disabling the register gate, the in-transaction count, and the `/info` computation each fail the matching tests and nothing else.

Two things stated rather than papered over:

- Nothing proves `Register` *picks* `CreateFirst`; swapping it back to `Create` leaves every test green, because the difference only shows under a genuine interleaving. A concurrent test would pass by luck as often as not. Noted in the test file.
- No Playwright case: every mode but `open` needs a differently-configured backend, and the suite runs against one shared stack.

Fixed on the way past: `TestRegister_lockedDBReturns503` passed only because another test had set the `config.C` global first — it nil-panicked when run alone. Both harnesses now set it per test.

## Verification

Containers, not just unit tests:

- closed → `registration_open: false`, register 403, **0 users**, 873 exercises seeded
- first-user → true, owner registers 201, flips to false, next signup 403, owner can still log in
- upgrade over a volume holding a demo account → warning logged, account intact
- `REGISTRATION=frst-user` → exits 1 with the valid set

Suites: Go `./...`, web type-check/lint/unit, shared 198, mobile 33, Playwright **93 passed / 1 skipped**. Web checked at 390px — link hidden on `/login`, `/register` explains itself.


.github/workflows/ci.yml
```diff
@@ -175,6 +175,10 @@ jobs:
         run: |
           echo "JWT_SECRET=ci-test-secret-$(openssl rand -hex 16)" >> .env
           echo "PORT=80" >> .env
+          # Demo seeding is off by default now, but the suite needs that account: the
```
Cawlumm: What we need to do is make it so suite doesnt need an existing account it makes its own to test those flows.

backend/config/config.go
```diff
@@ -69,3 +111,15 @@ func getEnv(key, fallback string) string {
 	}
 	return fallback
 }
+
+// getEnvBool reads a boolean env var. Anything unrecognised falls back rather than
+// being treated as false: DEMO_MODE=yes should not quietly mean "off".
+func getEnvBool(key string, fallback bool) bool {
```
Cawlumm, line 119: This is confusing, why not just have a standard instead of a complex method like this

Cawlumm:
## Convention check against other self-hosted apps

Compared the registration UX against Gitea, Vaultwarden, Immich, Lemmy, Mastodon, Uptime Kuma, Karakeep, Linkwarden and wger before finalising.

**Already matching:** server-side enforcement (universal), advertising state on a public API ([Mastodon's `registrations.enabled`](https://docs.joinmastodon.org/entities/Instance/)), hiding the affordance when closed, defaulting open ([wger's `ALLOW_REGISTRATION`](https://github.com/wger-project/wger/issues/402) defaults True), and a tri-state enum ([Lemmy's `registration_mode`](https://github.com/LemmyNet/lemmy/issues/3095)).

Worth noting how often peers ship this as UI-only and get it wrong: [Linkwarden #1485](https://github.com/linkwarden/linkwarden/issues/1485), [Karakeep #1830](https://github.com/karakeep-app/karakeep/issues/1830), [Gitea #36115](https://github.com/go-gitea/gitea/issues/36115). [Vaultwarden](https://github.com/dani-garcia/vaultwarden/wiki/Disable-registration-of-new-users) deliberately shows a button that errors, but only because the upstream web-vault is a static asset it cannot modify — there is [an open request to hide it](https://vaultwarden.discourse.group/t/hide-registration-form-when-signups-allowed-false/2205).

**Two changes came out of it** (2eaa61c):

1. A closed login screen now says *"This server is not accepting new accounts."* rather than showing nothing. Hiding is the convention, but [mastodon/mastodon#21556](https://github.com/mastodon/mastodon/issues/21556) documents that removing it without a trace confuses people who came specifically to sign up. Nothing clickable, so still no path that 403s.

2. `first-user` is now documented as a setup mode to move off, with a startup notice while the door is open.

   The reason is [immich-app/immich#24479](https://github.com/immich-app/immich/issues/24479): Immich's equivalent mode reopened admin registration when the data volume had not mounted at boot, letting anyone claim the instance. Closed as not planned. Ours is arguably easier to hit — a `docker compose up` from the wrong directory silently creates an empty `./data`.

   No in-database marker can guard against this, since it would be missing for the same reason. `REGISTRATION=closed` lives in `.env` and survives, so the docs route people there, and the backend now logs on every start where `first-user` is live with an empty table.

`first-user` was kept rather than dropped because the bootstrap problem it solves is real for anyone running `docker compose up -d` on a VPS with DNS already pointed at it — the case the issue was written around. Uptime Kuma, Immich and Gitea all do a version of it; wger documents doing it by hand.

backend/.env.example
```diff
@@ -17,6 +17,15 @@ DB_PATH=./data/lyftr.db
 JWT_SECRET=change-me-in-production-min-32-chars!!
 JWT_EXPIRY=3600
 
+# Registration — open | first-user | closed. See the root .env.example for what each
```
Cawlumm: Should we be having comments on the .env example since the docs already explain what they are for.

.env.example
```diff
@@ -5,6 +5,22 @@ JWT_SECRET=change-me-to-a-random-32-plus-character-string
...
+# Who may create an account:
+#   open        anyone (default)
+#   first-user  only while no account exists — claim your instance on first visit
+#   closed      nobody; /api/v1/auth/register always returns 403
+# On a public hostname: start at first-user, then switch to closed once you have signed
+# up. first-user asks whether the users table is empty, and that table is on your data
+# volume — a volume that fails to mount looks like a fresh install and opens the door
+# again. Only closed, which lives here in .env, survives that. See the docs.
+# Takes effect on container restart; no rebuild needed.
+REGISTRATION=open
```
Cawlumm, line 9: do we have two env examples?

docker-compose.yml
```diff
@@ -14,6 +14,10 @@ services:
       - CORS_ORIGIN=${CORS_ORIGIN:-http://localhost}
       - DB_TYPE=sqlite
       - DB_PATH=/app/data/lyftr.db
+      # open | closed | first-user. See site docs → Configuration.
```
Cawlumm: do we need comments here again

fly.toml
```diff
@@ -14,6 +14,9 @@ primary_region = "fra"
   PORT       = "3000"
   DB_PATH    = "/app/data/lyftr.db"
   CORS_ORIGIN = "https://lyftr-demo.fly.dev"
+  # This is the public demo — the whole point is the shared demo@lyftr.local account,
```
Cawlumm: same here do we need comments

.github/workflows/ci.yml
Cawlumm: Agreed, and it turned out to be mostly true already — `e2e/fixtures.ts` registers a fresh account per worker, and the only spec naming `demo@lyftr.local` was the wrong-password case, which passes against any address because the server answers the same for a missing account.

What actually depended on the demo user was this gate. Moved it into the suite as `e2e/globalSetup.ts`: waits for `/info`, registers a throwaway account, polls the exercise library through it, and records the token so `globalTeardown` sweeps it. That covers what the old gate covered (API up + async seed finished) with nothing pre-existing.

So `DEMO_MODE=true` is gone from this job and CI now runs against the production default — an instance with no demo account. Verified that way locally: fresh volume, `docker compose up`, confirmed the demo login fails, then 93 passed / 1 skipped, teardown swept 5/5 including the probe. (cd8a19f)

backend/config/config.go
Cawlumm: Fair — replaced with `strconv.ParseBool`, so the standard library decides what a bool looks like and `yes`/`on` stop being special.

Kept one deviation: unparseable falls back to the default rather than reading as false. `DEMO_MODE=ture` turning seeding silently off is the failure this defends against. Test covers both. (cd8a19f)

backend/.env.example
Cawlumm: Trimmed. Left `# open | first-user | closed` and one line saying the demo password is public — the value set, and the one fact that changes whether you should set it. Everything else was the docs restated. (cd8a19f)

.env.example
Cawlumm: Yes, two, and they are for different things: the root one is read by `docker compose` (self-hosters), `backend/.env.example` is for `go run main.go` (contributors, per CONTRIBUTING.md). They overlap on `JWT_SECRET`/`PORT` and diverge elsewhere — the backend one carries `DB_*` and `EXERCISEDB_KEY`, the root one carries `BACKEND_ORIGIN`, which only means anything to the compose frontend.

Not consolidating here since it predates this PR, but worth an issue if the duplication bothers you. Trimmed both to a pointer plus the one non-obvious warning. (cd8a19f)

docker-compose.yml
Cawlumm: Removed — the value names are the documentation, and `${REGISTRATION:-open}` already shows the default. (cd8a19f)

fly.toml
Cawlumm: Cut to one line: `# Without this the demo deploys with no account to log into.`

Kept something here because this is the one place the value is non-obvious — `DEMO_MODE = "true"` in a production config reads like a mistake, and the next person to tidy it would break the public demo. (cd8a19f)

.github/workflows/ci.yml
Cawlumm: Done in cd8a19f. `web/e2e/globalSetup.ts` now registers its own throwaway account and polls `/exercises?limit=1` through it, so readiness no longer depends on a seeded user. Both the `DEMO_MODE=true` export and the "Wait for demo user to be seeded" step are gone from this job — it runs the production default now, which is also what a self-hoster gets.

backend/.env.example
Cawlumm: Agreed — stripped in cd8a19f. The docs are the single explanation; the example file just lists the vars.

docker-compose.yml
Cawlumm: Stripped in cd8a19f, same reasoning.

fly.toml
Cawlumm: Stripped in cd8a19f. `DEMO_MODE = "true"` stands on its own — the public demo needs the account, and that is the only reason it is set here rather than in `docker-compose.yml`.

PR #114, opened 2026-08-17
Change your password in-app, and reset it from the CLI when it's lost

Cawlumm:
## Why

Settings had **Sign out** and **Delete account** but no way to rotate a credential. The account could be destroyed from the app and not secured from it. Mobile had the same asymmetry.

Peer self-hosted apps mostly ship this web-only — the Nextcloud Android app still doesn't have it ([nextcloud/android#10539](https://github.com/nextcloud/android/issues/10539)), and the complaint in that thread is a user hunting the setting on their phone. Lyftr is mobile-first, so both clients get it here.

## The part that isn't obvious

Tokens are stateless. `ValidateToken` never touched the database and refresh tokens live **30 days**. A password change on its own would revoke nothing — a stolen refresh token would keep minting access tokens for a month, which is the exact threat changing a password is meant to answer.

So `users.token_version` is carried in every token as the `ver` claim and checked **on refresh only**. The per-request auth path stays free of database work. Other devices stop when their access token lapses (≤1h) and their refresh is already dead; the device that made the change gets a fresh pair in the response and stays signed in.

**No one is signed out by the upgrade.** Tokens minted before the claim existed decode as `0`, accounts have always started at `1`, and `NormalizeTokenVersion` folds the two together. Those legacy tokens die the moment the account moves to `2`.

**Concurrent changes.** The store's `UPDATE` is conditional on the hash the handler verified. Two changes racing off the same hash would otherwise both write — the loser's password silently winning while its user was told theirs had been set — and both would land on `version+1` rather than `+2`, leaving the first change's tokens alive.

## Bug this surfaced

The shared client's 401 interceptor exempted only `/auth/`. A wrong current password 401s on `/me/password`, so it read as an expired session: refresh, retry, fail, sign out. **A typo cost you the session.** `/me/password` is a credential check and now says so. Pinned by a shared unit test and an e2e spec.

## Also

- Rejects reusing the current password **before** hashing — otherwise it burns a bcrypt round and still bumps `token_version`, signing every other device out for a change that never happened.
- `min=8` mirrors `RegisterRequest`, so a weaker password can't be *set* than can be *registered*.
- Looks the user up by ID, not by the email in the token.
- Docs: `configuration.md` gains "Changing your password", including that there is **no reset** and what DB-level recovery actually means.

## Verification

| | |
|---|---|
| Go | all packages pass, 9 new tests in `password_test.go` |
| Shared | 200 pass (2 new) |
| Web | type-check, lint 0 errors, 80 unit, build |
| Mobile | type-check, 33 jest, lint 0 errors |
| Site | builds |
| **Playwright** | **96 passed / 0 failed**, incl. 2 new specs |

E2E ran against an isolated backend on `:3099` and vite on `:5199` (`BASE_URL` + `API_URL` overrides), so it did not depend on any local dev stack.

## Known gaps

- **Mobile is not device-verified.** Type-checked and lint-clean; a Maestro flow was written for it but is **not in this diff** — `mobile/.maestro/` is gitignored and no flow in that directory is tracked today. Worth un-ignoring separately.
- **No password reset**, only change. An instance with no mail server can't send a link; recovery is documented as a DB operation instead.
- **`DELETE /me` still takes no password** — session theft is enough to delete an account. Pre-existing, and its own fix.


Cawlumm:
Added `reset-password` after checking what the rest of the space does.

The recovery path documented earlier in this PR was a `sqlite3` UPDATE against an `htpasswd`-generated hash. It works — I verified it end to end — but no comparable project asks that of a user:

| App | Command | Password input | Sessions |
|---|---|---|---|
| Immich | `immich-admin reset-admin-password` | prompt | **asks to invalidate** |
| Nextcloud | `occ user:resetpassword <user>` | prompt | — |
| wger / Paperless | `manage.py changepassword <user>` | prompt | — |
| Grafana | `grafana-cli admin reset-admin-password` | arg optional | — |
| Gitea | `admin change-password --username X --password Y` | **argument** | — |

So:

```bash
docker compose exec backend ./lyftr-api reset-password you@example.com
```

Prompts twice with no echo. **No `--password` flag** — Gitea's is the outlier and it leaves the password in shell history and the container's process list. Piped stdin still works for unattended use.

Sessions are invalidated via the `token_version` this PR already adds. Immich is the only peer that handles this and it asks; here it's unconditional, since an operator running a reset is usually evicting someone.

`golang.org/x/term` was already in the module graph via x/crypto, so this promotes an indirect requirement rather than adding a dependency — pinned to the already-resolved v0.42.0, since taking latest also dragged x/sys from v0.43.0 to v0.47.0.

Verified against a live instance: unknown email rejected, short password rejected, and after a reset the old password 401s, the new one 200s, a pre-reset refresh token 401s, and the account's data is intact. Plus three store-level tests.

backend/controllers/auth.go
```diff
@@ -129,6 +131,71 @@ func (h *Handler) Login(c *gin.Context) {
 	utils.OK(c, models.AuthResponse{Token: access, RefreshToken: refresh, User: user})
 }
 
+// ChangePassword sets a new password for the signed-in caller and ends every other
+// session. The device that made the change gets a fresh token pair in the response, so
+// it stays signed in while the rest fall off as their access tokens lapse.
+func (h *Handler) ChangePassword(c *gin.Context) {
```
Cawlumm, line 159: We are making sure that when it ends the session when the password changes, it is just for that user correct. Not other users on app?

backend/stores/user.go
```diff
@@ -23,11 +23,69 @@ func (s *UserStore) GetMe(uid int64) (models.User, error) {
...
-	).Scan(&u.ID, &u.Email, &u.Password, &u.CreatedAt, &u.UpdatedAt)
+		`SELECT id, email, password_hash, token_version, created_at, updated_at FROM users WHERE email = ?`, email,
+	).Scan(&u.ID, &u.Email, &u.Password, &u.TokenVersion, &u.CreatedAt, &u.UpdatedAt)
 	return u, err
 }
 
+// GetByID is GetByEmail for an already-authenticated caller: it carries the hash, so
+// the password-change handler can verify the current password without trusting the
+// email in the token. sql.ErrNoRows if the account was deleted mid-session.
+func (s *UserStore) GetByID(uid int64) (models.User, error) {
```
Cawlumm, line 34: Make sure that all of these functions have guard clauses with null/empty checking if not already there

backend/stores/user.go
```diff
@@ -126,6 +184,48 @@ func createUserTx(tx *sql.Tx, email, hash string) (int64, error) {
...
+// ResetPassword is the operator's way in, for the account whose password is lost. Unlike
+// ChangePassword there is no old hash to verify against — the whole point is that nobody
+// knows it — so this is keyed on email and guarded only by having a shell on the server.
+//
+// It bumps token_version for the same reason the in-app change does, and here it matters
+// more: an operator resetting a password may be doing it because someone else got in, and
+// a new password is worthless while the intruder's refresh token still mints access
+// tokens for the rest of its 30 days.
+func (s *UserStore) ResetPassword(email, newHash string) error {
+	res, err := s.db.Exec(
```
Cawlumm, line 225: So this reset password is executed through the cli correct?

backend/go.sum
```diff
@@ -101,6 +101,8 @@ golang.org/x/sync v0.20.0/go.mod h1:9xrNwdLfx4jkKbNva9FpL6vEN7evnE43NNNJQ2LF3+0=
 golang.org/x/sys v0.6.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.43.0 h1:Rlag2XtaFTxp19wS8MXlJwTvoh8ArU6ezoyFsMyCTNI=
 golang.org/x/sys v0.43.0/go.mod h1:4GL1E5IUh+htKOUEOaiffhrAeqysfVGipDYzABqnCmw=
+golang.org/x/term v0.42.0 h1:UiKe+zDFmJobeJ5ggPwOshJIVt6/Ft0rcfrXZDLWAWY=
```
Cawlumm, line 104: Make sure with these new dependencies we are not bringing in any vulnerabilities

backend/go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/golang-jwt/jwt/v5 v5.3.1
 	github.com/joho/godotenv v1.5.1
 	golang.org/x/crypto v0.50.0
+	golang.org/x/term v0.42.0
```
Cawlumm, line 12: Same as https://github.com/Cawlumm/lyftr/pull/114/changes#r3803530668

backend/main.go
```diff
@@ -43,6 +43,12 @@ func warnIfUnclaimed() {
 }
 
 func main() {
+	// Subcommands are matched before flag.Parse, which would otherwise treat
+	// "reset-password" as a positional argument and start the server as usual.
+	if len(os.Args) > 1 && os.Args[1] == "reset-password" {
```
Cawlumm: Is this the cleanest way to handle this logic through cli, the cli approach is correct but our logic to do it I am questioning.

backend/resetpassword.go
```diff
@@ -0,0 +1,125 @@
...
+// whoever has a shell on the box.
+//
+// The password is prompted for, never taken as a flag. A --password argument is visible in
+// `ps`, in shell history, and in any process-listing the container exports, which is why
+// Immich, Nextcloud, wger and Paperless all prompt. Gitea's --password flag is the odd one
+// out and carries exactly those problems.
+//
+// Returns the process exit code.
+func runResetPassword(args []string) int {
+	if len(args) != 1 || strings.HasPrefix(args[0], "-") {
```
Cawlumm, line 43: Should we be doing print statements in prd? We should be logging instead right, print statements if they are the same as sout in java are not good in prd.

backend/controllers/auth.go
Cawlumm: Correct — one user only, and now pinned by a test rather than left to inspection.

The UPDATE is `WHERE id = ?` on the single row, and `token_version` lives on `users`, so there is no shared state to spill. Added `TestChangePasswordLeavesOtherAccountsAlone`: user A changes their password, and user B's hash, `token_version`, refresh token and password are all asserted unchanged. `TestResetPasswordLeavesOtherAccountsAlone` already covered the CLI side.

backend/stores/user.go
Cawlumm: Added where an empty value is actually dangerous, deliberately not everywhere.

The two statements that **write** a password (`ChangePassword`, `ResetPassword`) now return `ErrEmptyHash`. An empty hash is uniquely bad: it stores fine, then fails bcrypt against every password, so the account is silently locked out — and `token_version` still gets bumped, ending the sessions too.

Not added to the readers (`GetMe`, `GetByID`, `TokenVersion`, `GetByEmail`, `FindEmailFold`, `Count`). An empty email or zero id there matches no rows and comes back as `sql.ErrNoRows`, which every caller already handles. A guard would restate what the query does. No other store in the repo guards this way either — the `len(...) == 0` checks in `program.go` are about empty `IN` clauses, which is malformed SQL, a different problem.

Test: `TestPasswordWritesRefuseAnEmptyHash`.

backend/stores/user.go
Cawlumm: Yes — CLI only. There is no HTTP route to it.

`docker compose exec backend ./lyftr-api reset-password you@example.com`

It is keyed on email with no old-password check, which is only safe because reaching it requires a shell on the server. The in-app path (`PUT /me/password` → `ChangePassword`) is the one that verifies the current password. Deliberately kept apart so the unauthenticated-by-design variant can never be reached over the network.

backend/go.sum
Cawlumm: Ran `govulncheck` rather than eyeball it. **`golang.org/x/term` contributes zero** — it does not appear anywhere in the report.

It also is not really a new dependency: it was already in the module graph via `x/crypto`. This promotes an indirect requirement to a direct one, pinned to the version already resolved (v0.42.0). I did not take latest on purpose — `go get golang.org/x/term` without a version also dragged `x/sys` from v0.43.0 to v0.47.0 for no reason.

The scan does report 18 findings, all pre-existing and none from this PR:

- **17 are Go standard library**, found in `go1.26.1` — my local toolchain. The Dockerfile builds on `golang:1.26-alpine`, a floating tag, so released images already pick up the newest 1.26.x patch. Worth bumping `go.mod`'s `go 1.26.1` directive anyway so local builds match.
- **1 is `github.com/quic-go/quic-go@v0.59.0`**, pulled in transitively by `gin` via `http3` (`go mod why` confirms the chain).
- `golang.org/x/net@v0.52.0` shows up too but `go mod why` reports the main module does not need the package.

None of it belongs in a password-change PR. Happy to open a separate dependency-bump PR — say the word.

backend/go.mod
Cawlumm: Same answer as the `go.sum` thread: `x/term` adds no vulnerabilities, and it was already in the graph via `x/crypto` — this promotes it to a direct requirement, pinned to the already-resolved v0.42.0 so nothing else moves.

backend/main.go
Cawlumm: Good instinct — it was hiding a real bug, now fixed in `ce6074c`.

The old single-string check meant anything that was **not** `reset-password` fell through to `flag.Parse` and started the server. So:

```
lyftr-api reset-passwrod you@example.com
```

booted the API, looked like it worked, and left you believing a password had been reset. On a typo, during recovery, when you are least able to notice.

Now a table plus an explicit reject:

```go
var subcommands = map[string]func(args []string) int{
	"reset-password": runResetPassword,
}
```

`dispatchSubcommand` returns the exit code, or `-1` meaning "carry on and start the server". A bare word not in the table exits 2 and names what does exist; anything starting with `-` is still handed to `flag.Parse`, so `-version` is untouched. Adding a second command is one map entry.

Verified:

```
$ lyftr-api reset-passwrod foo@bar.com
unknown command "reset-passwrod"
available commands: reset-password
$ echo $?
2
```

Covered by `main_test.go` — table-driven over no-args, `-version`, `--version`, and two unknown words.

backend/resetpassword.go
Cawlumm: Pushing back on this one, though the underlying worry is right for server code.

`log` and `fmt` are not competing here — they go to different places for different readers. This is a CLI, and stdout **is** its interface, the same as `git`, `kubectl` or `gh`. Routing prompts through `log` would also stamp them, giving you `2026/08/18 09:14:22 New password: ` on the prompt line.

What matters is the stream split, and that is already right — verified:

```
$ lyftr-api reset-password you@example.com 2>/dev/null
Password reset for you@example.com.
Every existing session for that account has been signed out.

$ lyftr-api reset-password nobody@x.com 2>&1 1>/dev/null
no account found for nobody@x.com
```

Results on stdout, errors on stderr, exit 0/1/2. Errors never appear on stdout, so piping the output somewhere is safe.

The server side is unchanged and still uses `log` throughout — `config.Load()` and `db.Connect()` keep logging migrations and warnings, and Go sends those to stderr, so they never contaminate the CLI's stdout either. Nothing here is `System.out` in a long-running service; the process exists to print one result and exit.

PR #122, opened 2026-08-20
Refresh the cache by id, and stop the picker double-fetching its first page

Cawlumm:
Refreshing asked open-exercise-db for its **entire export** and then discarded
everything it wasn't already holding. That was the only route available — nothing
could fetch a named set of rows — so re-reading thirty exercises cost a 300 KB
response.

oedb now has an `ids` filter (Cawlumm/open-exercise-db#21), so refresh asks for
exactly the rows it holds, in batches of 100.

**Measured against production**, 235 cached exercises:

| | before | after |
|---|---|---|
| requests | 1 export | 3 batched |
| transferred | ~300 KB | proportional to what's held |
| time | — | **3.1s** |
| export calls | 1 | **0** |

## Details worth reviewing

- **Batches are sequential, not concurrent.** There are few of them, each response
  is cached upstream, and a burst of parallel requests is exactly the shape that
  trips a rate limiter for a wall-clock saving not worth having.
- **`per_page` is set to the batch size** on every request. Without it the response
  comes back paginated at the default and the tail of each batch goes missing —
  which surfaces as "some rows never seem to update", not as an error.
- **Empty cache makes no request at all**, rather than reaching for the catalog to
  discover it has nothing to refresh.

## Verification

The test asserts on the *request*, not just the result — "it refreshed the right
rows" was true of the old implementation too. Pointing refresh back at the export
fails it with:

```
refresh fetched /api/v1/exercises/export without naming ids
refresh pulled the export; it should ask for the ids it holds
```

Batching is covered at the boundary (250 ids → 3 requests, none over 100, each
with a matching `per_page`), since that's where rows would be lost silently.

`stores` and `oedb` pass locally; CI runs the rest.

### DIFF
--- a/backend/oedb/client.go
+++ b/backend/oedb/client.go
@@ -446,3 +446,48 @@ func (c *Client) Export(ctx context.Context) ([]Exercise, error) {
 	}
 	return out.Exercises, nil
 }
+
+// maxIDsPerRequest is oedb's documented cap for the ids filter. Exceeding it is
+// a 422, not a truncation, so the batching below has to respect it rather than
+// discover it.
+const maxIDsPerRequest = 100
+
+// ListByIDs fetches specific exercises by their oedb UUIDs.
+//
+// This is what refreshing a cache should cost: a request proportional to what is
+// held, rather than the full catalog export filtered down afterwards. Until oedb
+// grew this filter the export was the only route, which meant ~300 KB to
+// re-read a handful of rows.
+//
+// Requests are issued one batch at a time rather than concurrently. The batches
+// are few, each response is cached upstream, and a burst of parallel requests is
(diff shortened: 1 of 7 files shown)


Cawlumm:
Status check as part of a PR sweep — no code changes here.

Green (14 checks pass, 4 skipped) and mergeable, but **6 commits behind `main`**, so it
wants `git merge origin/main` before it lands rather than a merge on a base it has not seen.
Not rebased, per the repo's no-force-push rule.

Left alone deliberately: the branch is checked out in a local worktree with uncommitted
changes across `exercises.go`, `stores/exercise.go`, `stores/exercise_test.go`,
`ExercisePicker.tsx` and `WorkoutExercisePicker.tsx` — so whoever has that work in progress
should be the one to bring it up to date, not a background sweep.

PR #134, opened 2026-08-21
Let a saved food be removed from My Foods

### LINKED ISSUE #115: Delete Foods from 'My Foods'
I added the same food twice to my 'My Foods' list thinking that I could specify different amounts. What I ended up with was a duplicate entry that I am unable to get rid of to my knowledge.

I would expect an elipses or trash can to display an icon or label to remove an item from my personal list, but I am not finding it. Is it possible to remove entries from my list, and if so, how?

Thank you!

Image

Image

Cawlumm:
Closes #115

`DELETE /api/v1/food/saved/:id` has existed since saved foods shipped — user-scoped, 404s on a miss, covered by `TestDeleteSavedFood_success` and `TestDeleteSavedFood_ownershipEnforced`. `savedFoodsAPI.delete` has been sitting in the shared client alongside `list` and `create`. Neither app ever called it.

So My Foods was a list you could only add to. @pdschneider hit exactly that in #115: two identical rows, no way to remove either. He went looking for "an ellipses or trash can" — the affordance genuinely wasn't there.

**Backend is untouched.** This is client-only.

## Web

Delete hangs off the row the way `Food.tsx` already does it: a danger `IconButton`, then an inline confirm strip that replaces the row.

That required restructuring `FoodResultRow` — it was a `<button>` wrapping the whole row, and a `<button>` inside a `<button>` is invalid HTML that React warns about. The component now branches on `onDelete`: without it, it renders exactly the markup it always did. Recent and search results are the hot path on this screen and are not what's being changed here.

## Mobile

Follows `FoodEntryRow`, the nearest neighbour in the same screen family — a `⋮` opening an `ActionSheet`, with delete routed through the shared `ConfirmSheet` via `deleteConfirmProps`. No new primitives; `ActionSheet` already defers its action until after the dismiss animation so the confirm sheet isn't presented mid-close.

The row moves to `mobile/src/components/nutrition/FoodResultRow.tsx` so the delete path is testable without standing up the whole log flow.

## Failures aren't swallowed

`FoodEntryRow` currently discards a failed delete. Copying that here would leave the row on screen with no explanation — indistinguishable from a dead button, which is a poor thing to add to a report about a missing affordance. Web surfaces it inline; mobile surfaces it through the existing `Alert`.

## Tests

- **Web e2e** (`food.spec.ts`): creates its own saved food, deletes it through the UI, then **reloads and re-checks** — the row vanishing on its own only proves local state was filtered, not that the request reached the server.
- **Mobile unit** (6 cases): mostly guarding that Recent and search *didn't* grow a delete control, since the row is shared by all three tabs. Also covers the success and failure paths.

## Verification

| Gate | Result |
|---|---|
| `go test ./controllers/ ./db/ ./routes/` | pass |
| `shared:type-check` + `shared:test` | pass (224) |
| web type-check / lint / unit / build | pass (83 unit; lint at the 52-warning cap, no new ones) |
| mobile type-check / lint / test | pass (39) |
| Playwright, full suite | **99 passed, 1 skipped** |

## Deliberately not in this PR

@pdschneider was saving the same food twice *on purpose*, trying to keep two serving sizes. A unique constraint on `(user_id, name, brand)` would block that rather than help it. The actual gap is that there's no way to edit a saved food, which needs a `PATCH /food/saved/:id` that doesn't exist yet — filing separately.

### DIFF
--- a/mobile/app/(tabs)/nutrition/log.tsx
+++ b/mobile/app/(tabs)/nutrition/log.tsx
@@ -10,10 +10,11 @@ import {
   type FoodSearchResult, type SavedFood,
 } from '@lyftr/shared'
 import {
-  AppText, Button, Card, DateInput, IconButton, Label, NumberField,
+  Alert, AppText, Button, Card, DateInput, IconButton, Label, NumberField,
   NumericKeyboardAccessory, NUMERIC_ACCESSORY_ID, Screen, SearchField, SegmentedControl, Toggle,
 } from '../../../src/components/ui'
 import { BarcodeScanner } from '../../../src/components/nutrition/BarcodeScanner'
+import { FoodResultRow } from '../../../src/components/nutrition/FoodResultRow'
 import {
   MACRO_COLORS, MACRO_TEXT, MEALS, MEAL_COLORS, MEAL_ICONS, MEAL_LABELS, type Meal,
 } from '../../../src/components/nutrition/nutritionMeta'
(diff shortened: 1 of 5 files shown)


Cawlumm:
Superseded by #136, which contains this commit and replaces the UI it added. Unfavouriting is now one tap on a star, so the ⋮ menu this PR introduced is gone — a menu to remove something is three interactions for what a second tap undoes. Closing so `main` shows only the end state rather than a commit that adds a control the next one removes.

PR #136, opened 2026-08-21
Favorites: one star, tappable anywhere

### LINKED ISSUE #115: Delete Foods from 'My Foods'

Cawlumm:
Closes #115. Supersedes #134, now closed — this branch contains its commit and replaces the UI it added.

Started as "stop the duplicate in #115 being creatable", and testing it on a phone turned it into a rework of what favouriting *is*.

## The finding

Favouriting was welded to logging. The star only fired inside `handleLog`:

```ts
await foodAPI.log(payload)                     // payload = scaleServing(selected, servings)
if (saveToMyFoods) {
  await savedFoodsAPI.create({ name: selected.name, ... })   // ← `selected`, unscaled
}
```

So you could not favourite a food without also eating it, and could not favourite anything found by search at all. The portion goes on the log; the bare food goes on the favourite — which means @pdschneider's approach in #115 was never going to work. He starred the same food twice hoping to keep 100g and 170g, and servings is precisely what the star omits, so both rows came out identical by construction.

He was reaching for a **custom food**. What exists is a **star**. [MyFitnessPal](https://support.myfitnesspal.com/hc/en-us/articles/360032623471-How-do-I-add-an-item-to-My-Foods) and [Cronometer](https://support.cronometer.com/hc/en-us/articles/360018240312-Create-a-Custom-Food) keep those as separate concepts — custom foods are authored and editable, favourites are shortcuts and aren't.

## What changed

**The star is its own action.** One tap on, one tap off, on every row of all three tabs and beside the food on the detail step. No confirmation: a favourite is a bookmark, not a record, and a second tap restores it — which retires the mobile ⋮ → ActionSheet → ConfirmSheet, three interactions to undo something that now takes one.

**"My Foods" → "Favorites".** MFP's term means an editable personal database, which this isn't, and borrowing it is plausibly what set the expectation in #115. `Favorites` also carries the verb the rest of the copy needs — "Add to Favorites", "In your Favorites" — where `Saved` gave us "Save to Saved". Bookmark icon → star, filled when set. Internal names untouched: the tab value stays `myfoods`, the table stays `saved_foods`.

**`UNIQUE(user_id, name, brand)` + idempotent `CreateSaved`.** The client check reads a list fetched when the screen opened, so a second device or a stale list would sail past it. Starring something already starred returns the existing row with **200** instead of 201.

**Pull-to-refresh on the mobile log screen** — the only list screen without it; the other five already share one `RefreshControl` idiom.

## The migration deletes rows

`ensureIndex` is `log.Fatal` and `CREATE UNIQUE INDEX` fails against a table holding duplicates, so the dedupe runs first, behind a `migration_flags` entry, and **the index is only created when the dedupe reports the table clean**. A transient error on that one `DELETE` must not stop the server booting — least of all for the installs the migration exists to repair. Running without the index is much the smaller problem: the clients still check and `CreateSaved` still resolves conflicts.

Duplicates collapse to the **earliest** row. With no edit path anywhere, two rows sharing user/name/brand differ at most in macros from two search hits for the same product; either logs the same, and it's one tap to recreate.

## Adversarial pass

I attacked this rather than re-reading it. Four defects, all fixed here:

1. **Server wouldn't boot** on a failed dedupe (above).
2. **Appending the created row could duplicate it** — the 200 response can be a row the list already holds, after a refresh that raced the request; appending blind put two rows with the same React key in the list.
3. **One global in-flight flag made every other star dead** — a tap on any other row returned early, silently. Now keyed per food.
4. **Mobile's error banner sat inside the Favorites branch** — the star is on all three tabs now, so a failure starring a search result rendered nowhere.

Verified under load, not by inspection: **twelve concurrent stars of one food → one 201, eleven 200s, exactly one row.**

## Found but deliberately not fixed here

`CreateSavedFood` never normalises input, so `"Oats"`, `"Oats "` and `" Oats"` are three separate favourites that render identically, and `{"name":" "}` stores a blank row. That validation predates this branch — filed as **#137**, and it matters more now precisely because this PR promises a uniqueness that whitespace quietly defeats.

Starring from a diary entry is the one surface still missing a star — **#138**.

## Verification

| Gate | Result |
|---|---|
| `go test ./controllers/ ./db/ ./stores/ ./routes/` | pass |
| shared type-check + tests | pass (229) |
| web type-check / lint / unit / build | pass (83 unit; lint 52, at the cap, no new) |
| mobile type-check / lint / test | pass (40) |
| Playwright | 26/26 food · 71 chromium · 18 mobile |

New coverage: dedupe keeps the earliest row and leaves other brands and other users alone, then the index forbids a re-insert; idempotent create returns 200 and the same id; `findSavedFood` matches on trimmed-brand equivalence and does *not* fold case; the row renders no star without the callback, doesn't select the food when the star is tapped, and is inert mid-flight. The e2e case that matters: **star a food from the detail view, leave without logging, confirm it's in Favorites and the diary has no entry for it.**

## Why #134 is closed rather than merged first

#134 added a ⋮ menu on mobile and a trash + inline confirm on web. This PR deletes both: unfavouriting is one tap on the star, so a menu to remove something is three interactions for what a second tap undoes. Merging both would put a commit on `main` that adds a control the next commit removes.

What survives from #134 is structural rather than visual — the row extracted into `mobile/src/components/nutrition/`, and the web row restructured (outer div, inner button, sibling control) so a trailing control could exist at all. The star now sits where the trash did.

### DIFF
--- a/mobile/src/components/nutrition/FoodResultRow.test.tsx
+++ b/mobile/src/components/nutrition/FoodResultRow.test.tsx
@@ -31,14 +31,6 @@ const ADD = 'Add Chicken Breast to Favorites'
 const REMOVE = 'Remove Chicken Breast from Favorites'
 
 describe('FoodResultRow', () => {
-  // A list that doesn't pass onToggleFavorite gets no star at all. Nothing uses this
-  // today, but the prop is optional and the row must not render a dead control.
-  it('renders no star when the list does not offer favouriting', () => {
-    render(<FoodResultRow item={item()} onPress={() => {}} />)
-    expect(screen.queryByLabelText(ADD)).toBeNull()
-    expect(screen.queryByLabelText(REMOVE)).toBeNull()
-  })
-
   // The label carries the toggle state, because the fill that conveys it visually is
   // invisible to a screen reader.
   it('offers to add when the food is not favourited', () => {
--- a/mobile/src/components/nutrition/FoodResultRow.test.tsx
+++ b/mobile/src/components/nutrition/FoodResultRow.test.tsx
@@ -84,7 +84,7 @@ describe('FoodResultRow', () => {
     )
 
     fireEvent.press(screen.getByLabelText(OPTIONS_LABEL))
-    fireEvent.press(screen.getByText('Remove from My Foods'))
+    fireEvent.press(screen.getByText('Remove from Favorites'))
     // ActionSheet defers the action until after its dismiss animation so the confirm
     // sheet isn't presented mid-close.
     act(() => { jest.runOnlyPendingTimers() })
@@ -116,7 +116,7 @@ describe('FoodResultRow', () => {
     )
 
     fireEvent.press(screen.getByLabelText(OPTIONS_LABEL))
-    fireEvent.press(screen.getByText('Remove from My Foods'))
+    fireEvent.press(screen.getByText('Remove from Favorites'))
     act(() => { jest.runOnlyPendingTimers() })
     fireEvent.press(screen.getByText('Remove'))
 
--- a/mobile/src/components/nutrition/FoodResultRow.test.tsx
+++ b/mobile/src/components/nutrition/FoodResultRow.test.tsx
@@ -1,4 +1,4 @@
-import { act, fireEvent, render, screen, waitFor } from '@testing-library/react-native'
+import { fireEvent, render, screen } from '@testing-library/react-native'
 import type { FoodSearchResult } from '@lyftr/shared'
 
 // Same preamble as the programs component tests: expo-router can't load under bare jest,
@@ -12,16 +12,8 @@ jest.mock('@react-native-async-storage/async-storage', () =>
 jest.mock('react-native-safe-area-context', () => ({
   useSafeAreaInsets: () => ({ top: 0, bottom: 0, left: 0, right: 0 }),
 }))
-jest.mock('expo-haptics', () => ({
-  selectionAsync: jest.fn(() => Promise.resolve()),
-  notificationAsync: jest.fn(() => Promise.resolve()),
-  impactAsync: jest.fn(() => Promise.resolve()),
-  NotificationFeedbackType: { Success: 'success', Warning: 'warning', Error: 'error' },
-  ImpactFeedbackStyle: { Light: 'light', Medium: 'medium', Heavy: 'heavy' },
-}))
 
 import { FoodResultRow } from './FoodResultRow'
-import { client } from '../../lib/lyftr'
 
 const item = (overrides: Partial<FoodSearchResult> = {}): FoodSearchResult => ({
   name: 'Chicken Breast',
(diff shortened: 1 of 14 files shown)


mobile/src/components/nutrition/FoodResultRow.test.tsx
```diff
@@ -0,0 +1,88 @@
+import { fireEvent, render, screen } from '@testing-library/react-native'
+import type { FoodSearchResult } from '@lyftr/shared'
+
+// Same preamble as the programs component tests: expo-router can't load under bare jest,
+// and the ui barrel reaches it via useTheme → lib/lyftr.
+jest.mock('expo-router', () => ({
+  router: { push: jest.fn(), navigate: jest.fn(), replace: jest.fn() },
+}))
+jest.mock('@react-native-async-storage/async-storage', () =>
+  require('@react-native-async-storage/async-storage/jest/async-storage-mock')
```
Cawlumm: A `require()` style import is forbidden 
CI / Mobile + shared checks
Check warning on line R10
Check warning:
A `require()` style import is forbidden

mobile/src/components/nutrition/FoodResultRow.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+)
+jest.mock('react-native-safe-area-context', () => ({
+  useSafeAreaInsets: () => ({ top: 0, bottom: 0, left: 0, right: 0 }),
+}))
+
+import { FoodResultRow } from './FoodResultRow'
```
Cawlumm: Import in body of module; reorder to top
CI / Mobile + shared checks
Check warning on line R16
Check warning:
Import in body of module; reorder to top
CI / Mobile + shared checks

web/e2e/food.spec.ts
Cawlumm: Collapse annotationCheck warning on line R67Check warning: Unexpected any. Specify a different typeCI / Web unit testsView detailsCollapse annotationCheck warning on line R67Check warning: Unexpected any. Specify a different typeCI / Web unit testsView details
      e.name.startsWith('E2ELog-') || e.name.startsWith('E2ESave-') || e.name.startsWith('E2EEdit-')
    )
    for (const e of toDelete) {
    const sfList = await request.get(`${API}/food/saved`, { headers: h })
    const sfb = await sfList.json()
Comment view    const sfToDelete = (sfb.data ?? []).filter((s: any) => s.name.startsWith('E2ESave-'))Collapse annotationCheck warning on line R77Check warning: Unexpected any. Specify a different typeCI / Web unit testsView detailsCollapse annotationCheck warning on line R77Check warning: Unexpected any. Specify a different typeCI / Web unit tests

Issue #141, opened 2026-08-23
[Android APK] Cannot input decimal numbers and Rest Timer countdown is missing


Cabiny6:
**What happened?**

I am experiencing two separate issues while logging my workouts using the Android APK app:

1. **Decimal input restriction:** When attempting to log weights or values during an exercise, the input field does not allow decimal numbers (e.g., I cannot type `12.5` or `12,5`). It seems the virtual keyboard configuration or field properties are forcing an integer-only input.
2. **Missing Rest Timer:** After completing and checking a set, I cannot see any visual rest time countdown anywhere on the screen. I am unsure if this feature is missing from the Android client or if it fails to render on mobile app screen sizes.

**Steps to reproduce**

**For the decimal issue:**
1. Open the Lyftr Android APK app and start a workout.
2. Go to an exercise set input field (Weight).
3. Try to type a decimal value (like `22.5`). The app blocks the entry of a dot or comma.
**For the rest timer issue:**
1. Complete a set during a guided workout and check it off.
2. Look for the rest timer countdown on the active screen. No visual timer appears.

**Expected behavior**

- Exercise input fields in the Android app should accept floating-point/decimal numbers (using `inputType="numberDecimal"` or equivalent).
- A visible rest timer countdown should trigger on the app layout right after a set is marked as done.

**Environment**

- Deployment: Docker
- Browser: mobile (APK)
- OS: Android 17 (CP2A.260805.005)

Cawlumm:
Hey @Cabiny6 — thanks for this one. It was a real bug and a bad one: on a locale whose
keypad draws a comma, the weight field was **unenterable**, not merely awkward. There's a
test build from #147 if you have a few minutes to confirm it on your own phone and Android
version.

**Download:** https://github.com/Cawlumm/lyftr/actions/runs/32880108923 — the APK
(`lyftr-v0.0.0-dryrun.17.apk`, 44 MB) is in the *Artifacts* section at the bottom, inside a
zip. GitHub requires you to be signed in to download an artifact, and it expires on
**2026-09-08**. It's signed with the same key as the releases, so it installs over your
current app and keeps you signed in. If Android refuses with "app not installed", uninstall
the old one first — your data lives on your server, not on the phone.

### What was actually wrong

Both halves of your sentence — *"the app blocks the entry of a dot or comma"* — were true,
for two different reasons in the same field:

**The comma never reached the app.** Every numeric field filtered typed text through
[...]

PR #143, opened 2026-08-24
Stop the dev loop crashing: let Expo configure Metro for the monorepo

Cawlumm:
> **Branch name is stale.** `fix/141-decimal-input-and-rest-timer` is where this started —
> it carried the #141 decimal fix until that was carved out into #147 (merged). GitHub
> cannot rename the branch under an open PR without rewriting published history, which this
> repo does not do, so the name stays wrong and the diff is the truth: 11 files, all build
> and dev configuration, no app source.

The mobile dev loop kept dying at startup, three separate sessions this week:

```
Failed to start watch mode                      (metro-file-map/src/Watcher.js:161)
TypeError: Cannot read properties of undefined (reading 'getSha1')
```

The second line is NativeWind reading a file map that was never built. The first is the
cause: `metro.config.js` pointed `watchFolders` at the **workspace root**, so
`metro-file-map` crawled `.git`, `backend/data`, `.playwright-mcp` and every session
worktree under `.claude` — a full checkout each — and exceeded its watch timeout before
Metro served a byte.

## The fix is deletion

Since SDK 52, `expo/metro-config` detects a monorepo and configures it, and
[the docs say to delete the manual version](https://docs.expo.dev/guides/monorepos/) rather
than keep it in step. Ours was the pre-52 recipe. What Expo picks is narrower and correct:

| | watchFolders |
|---|---|
| ours | `[ workspace root ]` |
| Expo's | `[ node_modules, packages/shared, mobile, web ]` |

`nodeModulesPaths` comes out byte-identical, so nothing is lost. Verified on a device:
Metro starts with zero watch errors, the app bundles (3712 modules) and renders, and a
marker string added to `packages/shared` appears in the freshly built bundle — which is the
behaviour the manual config existed to guarantee.

## What else is here

**`babel-plugin-transform-import-meta` removed.** It sat alongside the Metro resolver
override, both claiming to fix zustand's ESM build leaking `import.meta` into the web
bundle. Exporting the web bundle four ways settles which one works:

| config | occurrences of `import.meta` |
|---|---|
| override + plugin | 0 |
| **override alone** | **0** |
| plugin alone | 2 |
| neither | 2 |

The override is load-bearing; the plugin never did anything. That table now lives in
`metro.config.js` so the next person doesn't re-derive it.

**ABIs are per-profile** (`app.config.js`, keyed on `EAS_BUILD_PROFILE`). Previously one
static list applied everywhere, which optimised the wrong artifact: production builds an AAB
and [Play generates per-device APKs from it](https://developer.android.com/guide/app-bundle),
so restricting ABIs there buys no download size and only drops devices — and
[Play's 64-bit rule is additive](https://developer.android.com/google/play/requirements/64-bit),
so `armeabi-v7a` was never disallowed. Meanwhile the universal APK that testers actually
side-load had dropped `armeabi-v7a` and kept `x86_64`. Now: omitted for production,
`armeabi-v7a/arm64-v8a/x86_64` for preview.

**Workflow hardening**, both flagged by CodeQL or by review:
- `expo/expo-github-action` and `softprops/action-gh-release` pinned to commits. The second
  matters more — it runs in the job holding `contents: write`.
- `EXPO_TOKEN` moved from job-level `env` to the two steps that need it; it was reaching
  `actions/cache`, `upload-artifact` and third-party release code.

**`eas.json`**: `production` declares `distribution: store` + `app-bundle` (it was
inheriting internal/APK), and the `channel` keys are gone — they are consumed by
`expo-updates`, which this project does not install, and produced a warning on every build.

**`expo.install.exclude`** for `eslint-config-expo`: `expo-doctor` failed on it expecting
`~10.0.0` against an installed `57.0.0`, and `expo install --fix` acts on that by
downgrading a linter 47 majors. Expo's manifest is wrong for a package that isn't native.
Doctor now reports 17/18, and the remaining failure is real (see below).

**Scripts**: one added, `doctor` (`npx expo-doctor@latest`). **README** rewritten for what the
flow now is.

## What I removed after review said so

A reviewer asked "do we need this at all" about three things, and was right about all three:

- **an `adb reverse` script** — Expo CLI already does this when it launches the app
- **a `dev_client` CI job**, and then `expo-dev-client` itself, the `development` profile, `build:dev`/`start:go`
  and eventually `devFlow.test.ts` with it. The justification was that Expo Go cannot apply a config
  plugin and #79's is native XML — true, and already covered: any EAS build runs prebuild, so
  the **preview APK** has the plugin, and the release workflow proves it with `aapt2 dump
  xmltree` on every release. What a dev client adds is JS iteration against native code
  nobody has added, and it charged rent for it — installing the package changes what
  `expo start` means for every contributor. The `development` profile predated this PR and
  was decorative anyway (`developmentClient: true`, no package installed)
- **dev-host buttons on the sign-in screen** — replaced with `EXPO_PUBLIC_API_URL`, then
  that was removed too: it fills the field only when nothing is saved, while the failure
  that actually hurts is a *stale* saved URL, and it introduced a way for a developer's LAN
  address to reach a release bundle

And one I added and then deleted: an `.easignore`. It **replaces** `.gitignore` rather than
extending it, and mine didn't re-list `/android` — which is 1.7 GB on a machine that has run
prebuild. It would have tried to upload that into every cloud build.

## Known and deliberately not fixed

`expo-doctor`'s remaining failure is genuine: `expo` and `jest-expo` are two patch versions
behind. Bumping them nests a second copy of `expo` under `mobile/node_modules` while the
root keeps the old one, which doctor rightly calls worse than the drift, and `npm dedupe`
does not resolve it. That needs a lockfile regeneration — a dependency change with its own
blast radius, not a developer-experience PR. It is also why `expo install --check` is not in
CI yet.

## Gates

| | |
|---|---|
| mobile | 44 |
| shared | 254 |
| web | 83 + build |
| type-check | clean, all three packages |
| lint | 0 errors |
| device | Metro clean, app renders, shared fast-refresh confirmed |

`devFlow.test.ts` is gone too. It read `eas.json`/`app.json` and asserted values written in
the same commit — a change detector with no independent oracle. Its one scanning test looked
for developer-machine addresses in source, which no longer exist, and could be disarmed by
appending a bare `// __DEV__` comment to a file.


Cabiny6:
Thanks for addressing this! I noticed one extra detail: in List mode, even typing a full stop (.) gets stripped/deleted as well. I've attached a quick video demonstrating this behaviour in case it's helpful for this PR or follow-up fixes.

https://github.com/user-attachments/assets/2a7480fb-c203-4157-adf9-b3e6b503f5e2

Cawlumm:
Splitting this. It had grown to four unrelated things.

**The decimal fix moved to #146**, on its own branch cut from `main` — that's the part @Cabiny6 is waiting on in #141, and it's been green since this morning. It carries only `sanitizeNumericInput` and its tests; nothing about build tooling.

**This PR keeps the dev-flow work**: the `eas.json` corrections, the `__DEV__` guards and `devFlow.test.ts`, the dev quick-fill in the server field, the removed Tailscale hostname, and the `dev_client` runner job.

Also corrected a claim I made in this PR and had wrong. I asserted Expo Go couldn't run this project and repointed `npm start` at `--dev-client` on that basis. Checked properly on the emulator:

- Expo Go 54.0.8 is installed and matches the SDK exactly
- Metro serves a Go-compatible manifest — `200`, `"runtimeVersion":"exposdk:54.0.0"`
- `expo-updates` isn't a dependency of this project at all
- Expo Go bundles the SDK, so camera/secure-store/font/localization/splash/router are all present

The evidence I'd relied on was a misread: *"The expo-updates system is disabled due to an invalid configuration"* is logged at `"level":"warn"` by Expo Go's **own** bundled expo-updates, because a local dev project has no update URL. Routine.

What actually breaks Metro is an `@expo/cli` bug — it reads a `fetch` body twice in the startup dependency check:

```
TypeError: Body is unusable: Body has already been read
  at getNativeModuleVersionsAsync (@expo/cli/src/api/getNativeModuleVersions.ts:47)
```

Metro dies before binding; the failure then looks like the project's fault. `EXPO_NO_DEPENDENCY_VALIDATION=1` gets past it. Reverted in 1f02978 and documented in the README, along with `--tunnel` for when the device can't reach Metro.

`expo-dev-client` stays as a **fallback**, not the default — Expo Go can't apply a config plugin, so testing `expo-network-security-config` (#79) still needs a dev build.

scripts/link-device.js
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env node
```
Cawlumm: Do we need this script at all? Seems redundant.

mobile/app.json
```diff
@@ -66,6 +66,17 @@
           "enable": true,
           "networkSecurityConfig": "./network_security_config.xml"
         }
+      ],
```
Cawlumm: Is this what will limit the size of the production build?

mobile/src/components/authui.tsx
```diff
@@ -218,6 +218,24 @@ export function AuthDivider() {
...
+// alias for the host machine — the emulator's own `localhost` is the emulator — and a
+// simulator on iOS shares the Mac's loopback, so the two platforms need different
+// addresses. Typing either by hand on a soft keyboard, every time an emulator is wiped,
+// is the sort of friction that ends with someone hardcoding a URL in the client.
+//
+// `__DEV__` is a compile-time constant: Metro substitutes `false` for release builds and
+// the minifier drops the branch, so neither these addresses nor the buttons exist in a
+// production bundle. It has to stay a literal `__DEV__` check for that to hold — reading
+// the same thing from a runtime value would leave the branch in the shipped bundle.
+const DEV_HOSTS = __DEV__
```
Cawlumm: Is this necessary or overcomplication. What does the research say.

mobile/src/components/authui.tsx
```diff
@@ -301,6 +319,27 @@ export function ServerRow() {
               fontSize: 14,
             }}
           />
+          {DEV_HOSTS.length > 0 ? (
```
Cawlumm: same as R231

.github/workflows/eas-build.yml
```diff
@@ -105,11 +113,102 @@ jobs:
...
+  # screen does not change it. That is why this is a manual dispatch and not a per-push
+  # job: rebuilding it after a JS change produces an APK that behaves identically to the
+  # one already on the device, for ~20 minutes of runner time. Expo builds the same
+  # assumption into `eas build:dev`, which reuses a build whose fingerprint still matches
+  # rather than making a new one.
+  #
+  # `--local` keeps it on this runner, so it does not touch the EAS cloud build quota —
+  # the documented way to keep building once that quota is spent. Credentials still come
+  # from EAS, so this is signed with the same key as everything else.
+  dev-client:
```
Cawlumm: Does the research lead to a seperate client build for dev? Do we even need this any longer since we use metro locally?

scripts/link-device.js
Cawlumm: You're right — deleted.

Expo CLI already runs `adb reverse` when it launches the app, so `npm run android` covers the
normal path and the script was solving a problem the tooling solves. What it actually covered
was the narrower case I kept hitting in automation: attaching to an app that's *already*
running, where nothing has run `adb reverse` for you. That's now two lines in the README
rather than a script plus a dependency-free adb resolver.

The one thing worth keeping from it is the symptom, because it isn't obvious: when the device
can't reach Metro you don't get an error, you get a splash screen that never advances.

mobile/app.json
Cawlumm: Partly — and the distinction matters, so let me be precise about which build it shrinks.

`buildArchs` controls which ABIs get compiled in. The **production** profile builds an
`app-bundle` (AAB), and Play generates per-device splits from that anyway, so a user's
download was never carrying all four ABIs on that path.

Where it does change the size is the **sideloadable APK** — the `preview` profile, which is
what gets attached to a GitHub Release and what @Cabiny6 installed. That one is universal: it
carries every ABI in a single file, and `x86` alone was ~47% of the download for code that
only ever runs on an emulator.

The real cost isn't size, it's reach: dropping `armeabi-v7a` drops 32-bit-only devices
entirely, from both the APK and the AAB. That's the call Nextcloud made and Play has required
64-bit since 2019, but it is a product decision rather than a build tweak, which is why it's
pinned by a test with the reasoning next to it.

mobile/src/components/authui.tsx
Cawlumm: Overcomplication — and the research is unambiguous, so these are deleted.

[Expo's answer](https://docs.expo.dev/guides/environment-variables/) to "which backend does
this machine talk to" is an `EXPO_PUBLIC_`-prefixed environment variable, statically inlined
into the bundle at build time. Not UI:

```bash
# mobile/.env.local  (already gitignored)
EXPO_PUBLIC_API_URL=http://10.0.2.2:3000
```

`createServerStore` now takes a `fallback` that mobile fills from
`process.env.EXPO_PUBLIC_API_URL`, so the field is pre-filled with no buttons and no list.
It's strictly better than what I wrote: it covers a **physical device on your LAN** and an
iOS simulator too, because each machine sets its own address rather than choosing from a
hardcoded pair. A URL saved in the app still wins — this only fills a blank. Documented in
`mobile/.env.example`.

For the record, the deleted code wasn't wrong about the risk it was managing: `__DEV__` is
substituted at transform time, so those addresses genuinely could not reach a release bundle,
and `devFlow.test.ts` still fails if a developer-machine address appears outside `__DEV__`
code. The env var is just the smaller way to the same place — 39 lines of UI down to one
argument.

mobile/src/components/authui.tsx
Cawlumm: Same answer as above — both the constant and the button row are gone, replaced by
`EXPO_PUBLIC_API_URL` from `.env.local`.

.github/workflows/eas-build.yml
Cawlumm: Deleted, along with its dispatch input.

The research says a development build is the right tool when Expo Go can't run your app —
[Expo positions it](https://docs.expo.dev/develop/development-builds/introduction/) as "your
own version of Expo Go where you are free to use any native libraries and change any native
configuration". The question is whether *we* are in that situation, and today we aren't: Expo
Go runs everything this app uses, which is how all of this week's device testing happened.

The one genuine gap is a config plugin — `expo-network-security-config` (#79) is native XML
that Expo Go cannot apply, so testing the user-CA/cleartext policy needs a dev client. But
that's reachable with `npm run build:dev` (EAS, `development` profile) when someone actually
needs it. A CI job is infrastructure waiting for a user.

It also would have shipped broken: the job pinned Node 20, and #149 has since moved this
workflow to 22 because eas-cli's dependency tree now requires it. On its first dispatch it
would have died in setup with the exact error #149 fixed. That's a decent argument on its own
for not carrying a job nobody runs — nothing exercises it, so nothing tells you it rotted.

Merged `main` into this branch in the same push, so the workflow is on 22 and the decimal fix
(#147) is back.

mobile/src/components/authui.tsx
Cawlumm: Correction, since this thread now reads out of date: the `EXPO_PUBLIC_API_URL` replacement
mentioned above is **also gone**.

It filled the server field only when nothing was saved — while the failure that actually
bites is a *stale* saved URL (an emulator still pointing at an old address, "Failed to load
programs"), which a fallback never touches. And it introduced a way for a developer's LAN
address to reach a release bundle, since `EXPO_PUBLIC_` values are inlined.

So the answer to your original question ended up being simpler than either version: nothing
replaces the buttons. You type the server URL once on the sign-in screen and it persists,
which is a first-class feature of the app because every user self-hosts.

mobile/src/components/authui.tsx
Cawlumm: Same correction as the thread above — the `EXPO_PUBLIC_API_URL` replacement was removed too. Nothing replaces the buttons; the server URL is typed once and persists.

Issue #144, opened 2026-08-24
Rest timer only runs in Gym Mode, but List is the default view


Cawlumm:
Split out of #141, where @Cabiny6 reported no rest countdown after checking off a set. Not a rendering failure — the feature genuinely isn't in the view they were using.

## What's actually true

```
<RestTimerBanner> rendered in:  GymModeWorkout.tsx only
startRest() called in:          GymModeWorkout.tsx only
```

Both platforms. List mode — `mobile/app/(tabs)/workouts/active.tsx` and `web/src/pages/ActiveWorkout.tsx` — has neither the `startRest()` call on set completion nor the banner. Inside Gym Mode it works correctly: a floating banner on the non-logging phases, a docked one on the logging phase.

The default is `workout_layout ?? 'list'`, so most people never see it.

## Why it reads as broken rather than absent

Settings offers **"Rest timer — auto-start a countdown between sets"** as a global toggle, and a default rest duration beside it. Neither mentions Gym Mode. So a user in the default view turns the timer on, completes a set, and nothing happens — the setting appears to be ignored.

That's the part worth fixing regardless of which way this goes: either the feature follows the setting, or the setting says where it applies.

## Options

**Wire it into List mode.** The machinery already exists and is platform-shared in spirit — `useWorkoutSession`'s rest state, `useRestTimer`, and `RestTimerBanner` with its `docked` variant. List mode would call `startRest()` from its set-completion handler and render the banner. The work is deciding where it sits in a scrolling list of exercise cards, which is a genuine layout question rather than a wiring one — Gym Mode could dock it because it shows one exercise at a time.

**Or scope the setting.** Leave the feature in Gym Mode and make the toggle say so, so nothing silently does nothing.

I'd do the first — the toggle is a promise — but the layout call should be made deliberately rather than by copying the docked variant into a list and seeing what happens.

## Notes

- Do both platforms together. Splitting them is how #115 happened: the two apps disagreed about what an action meant.
- `startRest` is already guarded by `r > 0`, so a zero default still means "off" without special-casing.

PR #147, opened 2026-08-25
Accept the separator the keyboard actually emits (#141)

### LINKED ISSUE #141: [Android APK] Cannot input decimal numbers and Rest Timer countdown is missing

Cawlumm:
Fixes #141.

@Cabiny6 could not enter a decimal weight on an Android phone — neither `12,5` nor `12.5`.
Both halves of that sentence are real, and they are two different bugs in one field.

**The comma was deleted.** Every numeric field filtered typed text through `[^0-9.]`.
Android's `decimal-pad` draws the *locale's* separator, so on a European locale the
character the keypad emits was the one being stripped — and those keypads often have no
full stop to fall back on.

**The dot died on the next line.** `ActiveExerciseCard`'s weight cell re-derived `value`
from a *number* on every keystroke: `"12."` → `Number("12.")` → `12` → redraw `"12"`, and
React Native pushes that back into the native field. The separator is gone before the
fraction digit can be typed. `ExerciseFormCard` had solved this with a `useNumericText`
buffer and even said why in a comment; this cell never got the same treatment — and it
backs **List view, the default layout**, which is the screen the report came from.

Sanitizing is mandatory rather than defensive: `ReactEditText` replaces Android's
`KeyListener` specifically to *"permit all keyboard input through"*, so `keyboardType`
constrains what is **drawn**, never what arrives.

## The rule

wger fixed the same report against their Flutter app
([wger-project/flutter#1147](https://github.com/wger-project/flutter/issues/1147), now
`lib/core/number_input.dart`), and their rule is the one adopted here:

> digits, plus ONE separator; `.` and `,` both count whatever the locale, because
> *"a numeric keyboard cannot be pinned to a locale"*.

One refinement on top: whitespace never counts, because `fr`, `ru`, `sv`, `pl`, `cs`, `fi`,
`nb` and `uk` group with a space, so a pasted `"12,50 kg"` would otherwise read 1250.

| input | result | |
|---|---|---|
| `12,5` | `12.5` | the reported bug |
| `12.5` | `12.5` | the keypad may emit either |
| `82,5,` | `82.5` | first separator wins |
| `12,50 kg` | `12.50` | whitespace is a *group* char, never a decimal |
| `1.234,5` pasted | `1.2345` | not second-guessed — see below |

First rather than last is load-bearing, not a tie-break. A `TextInput` re-sends the whole
field per keystroke and these cells have no `selectTextOnFocus`, so a tap lands *inside* a
prefilled `82,5` — preferring the last separator turns one keypress into `825`, a silent
10x that backspace cannot undo because the fraction digit has become an integer digit.

**Grouping is deliberately not guessed.** Telling a group character from a decimal one
needs the locale's own separator, and then the other character is grouping by definition —
which is what ICU-based parsers do (react-aria's `NumberParser`, Expensify's
`LocaleDigitUtils`). No locale is injected in this PR, so a branch that guessed would be a
heuristic no comparable app ships. wger accepts the same trade. When #146 injects the
locale, the replacement is the ICU rule, not a smarter guess.

It could not be done mid-type in any case: `"1,"` arrives with nothing after it, so a
grouping-aware rule works on paste and lies while typing. `1,200` is `1.2` by both routes.

## Three copies, one rule

`[^0-9.]` existed in three components, which is exactly how the first attempt at this fixed
`NumberField` while the default layout stayed broken. There is one implementation now, in
`@lyftr/shared`, and `numericFields.test.ts` fails if a fourth appears — it reads the
sources rather than rendering anything, because the bug was never in what a component does
with the value, it was in one component not asking.

## Scope

Display formatting is **not** in here. The field still redraws canonical, so a German user
types `12,5` and sees `12.5` — entry works, notation is not yet theirs. That is the
enhancement and it stays on #146, which this PR is deliberately carved out of.

The rest timer half of #141 is not a bug: the countdown exists only in Gym Mode and List is
the default. That the setting promises it without saying so is fair, and is tracked
separately as #144.

Also bumps this repo's APK workflow to Node 22 — `eas-cli`'s dependency tree now declares
`engines.node >= 22`, so the tester build failed before it started.

## Gates

| | |
|---|---|
| shared | 251 |
| mobile | 42 |
| web unit | 83 |
| Go | `./...` ok |
| Playwright E2E | 99 passed, 1 skipped |
| type-check / build / lint | clean, 0 errors |

### DIFF
--- a/.github/workflows/eas-build.yml
+++ b/.github/workflows/eas-build.yml
@@ -88,11 +88,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          # 22, not the 20 the other workflows use: eas-cli's own dependency tree
-          # (@oclif/plugin-autocomplete) declares engines.node >= 22, so installing the
-          # CLI fails outright on 20 - "Found incompatible module", before any build
-          # step runs. Only this workflow installs eas-cli, so only this one needs it.
-          node-version: '22'
+          node-version: '20'
           cache: 'npm'
           cache-dependency-path: package-lock.json
       - uses: expo/expo-github-action@v8
@@ -140,11 +136,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          # 22, not the 20 the other workflows use: eas-cli's own dependency tree
-          # (@oclif/plugin-autocomplete) declares engines.node >= 22, so installing the
-          # CLI fails outright on 20 - "Found incompatible module", before any build
-          # step runs. Only this workflow installs eas-cli, so only this one needs it.
-          node-version: '22'
+          node-version: '20'
           cache: 'npm'
           cache-dependency-path: package-lock.json
       # Local Android builds compile with Gradle on this runner, so the JDK is ours to
--- a/.github/workflows/eas-build.yml
+++ b/.github/workflows/eas-build.yml
@@ -88,7 +88,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          # 22, not the 20 the other workflows use: eas-cli's own dependency tree
+          # (@oclif/plugin-autocomplete) declares engines.node >= 22, so installing the
+          # CLI fails outright on 20 - "Found incompatible module", before any build
+          # step runs. Only this workflow installs eas-cli, so only this one needs it.
+          node-version: '22'
           cache: 'npm'
           cache-dependency-path: package-lock.json
       - uses: expo/expo-github-action@v8
@@ -136,7 +140,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          # 22, not the 20 the other workflows use: eas-cli's own dependency tree
+          # (@oclif/plugin-autocomplete) declares engines.node >= 22, so installing the
+          # CLI fails outright on 20 - "Found incompatible module", before any build
+          # step runs. Only this workflow installs eas-cli, so only this one needs it.
+          node-version: '22'
           cache: 'npm'
           cache-dependency-path: package-lock.json
       # Local Android builds compile with Gradle on this runner, so the JDK is ours to
--- a/packages/shared/src/utils/number.ts
+++ b/packages/shared/src/utils/number.ts
@@ -40,3 +40,59 @@ export function clampValue(raw: string | number, min = 0): number {
   const n = typeof raw === 'number' ? raw : Number(raw)
   return Number.isFinite(n) ? Math.max(min, n) : min
 }
+
+// Typed text -> the canonical form the app stores. Every numeric field on mobile calls
+// this; web needs none of it, because <input type="number"> lets the browser parse the
+// locale's own separator. React Native has no equivalent hook: ReactEditText replaces
+// Android's KeyListener specifically to "permit all keyboard input through", so
+// `keyboardType` constrains what is *drawn* and never what arrives.
+//
+// The separator is the whole bug (#141). Android's decimal-pad draws the *locale's*
+// decimal character - a comma across most of Europe, and on those keypads often with no
+// full stop at all - so stripping it as "not a digit" left those users unable to enter a
+// decimal weight by any route.
+//
+// The rule is wger's, who fixed the same report against their Flutter app
+// (wger-project/flutter#1147, lib/core/number_input.dart):
+//
+//     digits, plus ONE separator. A "." and a "," both count, whatever the locale,
+//     because "a numeric keyboard cannot be pinned to a locale" - the keypad may emit
+//     either. Everything else is dropped.
+//
+// Which separator wins, where we go one step further than they do:
+//   - one KIND present -> the FIRST. "12,5," is 12.5 and "12.5." is 12.5.
+//   - both kinds -> the LAST, because then it is unambiguous: "1.234,5" and "1,234.5"
+//     are both 1234.5 without having to know which locale produced them. wger keeps the
+//     first unconditionally, which reads a pasted "1.234,5" as 1.23.
+//
+// First-when-unambiguous is not a detail. A TextInput re-sends the WHOLE field on every
+// keystroke and the cursor can sit mid-string, so preferring the last separator turns one
+// keypress inside a prefilled "82,5" into "825" - a silent 10x that backspace cannot undo,
+// because the fraction digit has become an integer digit.
+//
+// Grouping is deliberately not detected. It cannot be, mid-type: "1," arrives with
+// nothing after it, so any rule that reads grouping works on paste and lies while typing,
+// which is the path people actually use. wger does not attempt it either. "1,200" is 1.2
+// in every locale and by both routes.
+//
+// 'numeric' keeps digits only, so a separator cannot sneak a fractional rep in sideways.
+export function sanitizeNumericInput(raw: string, mode: 'numeric' | 'decimal'): string {
+  if (mode !== 'decimal') return raw.replace(/[^0-9]/g, '')
+
+  // Whitespace never counts as a separator. fr, ru, sv, pl, cs, fi, nb and uk group with
+  // a space, so a trailing space or a pasted "12,50 kg" would otherwise read as 1250.
+  const bare = raw.replace(/\s/g, '')
+
+  const at: number[] = []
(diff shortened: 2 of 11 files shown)


packages/shared/src/utils/number.ts
```diff
@@ -40,3 +40,58 @@ export function clampValue(raw: string | number, min = 0): number {
...
+// is unaffected, and wger accepts the same trade. If the locale ever is injected, replace
+// this with the ICU rule rather than a smarter guess.
+//
+// It also cannot be done mid-type in any case: "1," arrives with nothing after it, so a
+// grouping-aware rule works on paste and lies while typing, which is the path people use.
+// "1,200" is therefore 1.2 in every locale and by both routes.
+//
+// 'numeric' keeps digits only, so a separator cannot sneak a fractional rep in sideways.
+export function sanitizeNumericInput(raw: string, mode: 'numeric' | 'decimal'): string {
+  if (mode !== 'decimal') return raw.replace(/[^0-9]/g, '')
```
Cawlumm: Is there a react method/component that exists that can be used in place of this. Or is this customer logic needed to support both web and mobile shared. Always prefer existing libraries/methods if they exist.

Cawlumm:
**Status: ready, deliberately not merged yet.** Waiting on @Cabiny6 to confirm the build on
his own phone and Android version before this lands — the emulator run below is good
evidence, not proof, and he is the one who can close the loop.

Verified on an Android emulator set to Portuguese, running this branch's JS through Expo Go
over Metro, on the reporter's own screen (List view, active session). The comparison is the
same keystrokes into the same field with only the component swapped:

| typed | `main` | this branch |
|---|---|---|
| `12,5` | **`125`** — the reported bug, reproduced | `12.5` |
| `12.5` | — | `12.5` |
| `82,5,` | — | `82.5`, not 825 |
| `1.234,5` | — | `1.2` (first separator wins, then `round1`) |
| reps `8,5` | — | `85` |

Values survived a full app kill and relaunch, and Gym Mode's `NumberField` takes `12,5` on
the same keypad. The pt-PT `decimal-pad` draws **both** `,` and `.`, which is consistent
with the report naming both.

Two things worth recording from that session:

**A caret defect exists and it is not from this PR.** Editing a *prefilled* cell scatters
characters — `12,5` typed into `275` produced `212.6`. I reverted only
`ActiveExerciseCard.tsx` to `main`, hard-reloaded, and reproduced the same scattering, so it
is the pre-existing controlled-value caret reset, not a regression. wger handles it by
re-filtering the text before the caret to recompute the offset. Worth its own issue.

**#148 got sharper.** Typing `1.234,5` stores **1.2345** while the field settles on **1.2**,
because `displayWeight` is `round1` and the buffer resyncs to the rounded value. Screen and
record disagree from one uninterrupted piece of typing — details on the issue.

.github/workflows/eas-build.yml
```diff
@@ -88,7 +88,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
```
Cawlumm, line 91: what the eas build job failing without this? And if So can we merge it in on its own MR?

Cawlumm: Yes, it was failing, and yes — split out as #149.

It fails in setup, before building anything. `expo/expo-github-action` installs eas-cli, and
eas-cli 22.4.0 pulls `@oclif/plugin-autocomplete@3.3.0`, which declares `engines.node >= 22`:

```
error @oclif/plugin-autocomplete@3.3.0: The engine "node" is incompatible with this
module. Expected version ">=22.0.0". Got "20.20.2"
```

Both runs are on record: [32874687733](https://github.com/Cawlumm/lyftr/actions/runs/32874687733)
died at that step on Node 20, [32874931009](https://github.com/Cawlumm/lyftr/actions/runs/32874931009)
went green through to the signed APK on 22 — that second one produced the tester build linked
on #141.

Removed from this branch in 0833298, so #147 is back to eight files that are all the decimal
rule and the fields using it. #149 is worth merging first and on its own: nothing in this repo
caused the break (eas-cli moved), and until it lands no APK can be built from `main` at all —
including from a `v*` tag, where it would fail *after* the tag was pushed.

packages/shared/src/utils/number.ts
Cawlumm: Fair challenge, and I went and checked rather than answering from memory. Short version:
**web already uses the platform and has none of this; on mobile there is no library that
runs on both our platforms.**

**Web is already library-first.** No custom logic ships there — `<input type="number">` lets
the browser parse the locale's own separator. This function exists for React Native only.

**React Native has no equivalent seam.** [`TextInput`](https://reactnative.dev/docs/textinput)
exposes no input formatter; `onKeyPress` is documented as firing *before* `onChange` with no
`preventDefault`, and `maxLength` is the only native filter. Android's `ReactEditText`
replaces the platform `KeyListener` specifically to *"permit all keyboard input through"*, so
`keyboardType` constrains what is drawn, never what arrives. Filtering in `onChangeText` is
the only hook that exists.

**The three candidate libraries, checked today:**

| package | last publish | verdict |
|---|---|---|
| [`@internationalized/number`](https://www.npmjs.com/package/@internationalized/number) (react-aria) | 2026-08-14 | actively maintained, correct — **but breaks on iOS** |
| [`react-native-currency-input`](https://www.npmjs.com/package/react-native-currency-input) | 2023-09 | ~3 years stale; currency-shaped, not a plain decimal field |
| [`@wwdrew/react-native-numeric-textinput`](https://www.npmjs.com/package/@wwdrew/react-native-numeric-textinput) | 2022-04 | 4 years stale; depends on the deprecated `intl` polyfill |

The first is the one worth wanting, so I checked why it can't be used. Its `NumberParser`
calls `Intl.NumberFormat#formatToParts` in four places — I unpacked 3.6.7 and grepped:

```
package/dist/private/NumberParser.mjs:216    symbolFormatter.formatToParts(-10000.111)
package/dist/private/NumberParser.mjs:217    symbolFormatter.formatToParts(10000.111)
package/dist/private/NumberParser.mjs:218    pluralNumbers.map(n => symbolFormatter.formatToParts(n))
package/dist/private/NumberParser.mjs:239    ...).formatToParts(0.001)
```

Hermes does not implement `formatToParts` on iOS — [facebook/hermes#1188](https://github.com/facebook/hermes/issues/1188)
is still open, and it was still listed as a partial-support gap as of March 2026. So adopting
it means also shipping `@formatjs/intl-numberformat` plus CLDR locale data as a polyfill, to
replace ~20 lines. That is a bigger dependency footprint and a bigger runtime cost than the
thing it removes.

**Why there is no built-in to fall back on either:** JavaScript has a number *formatter* and
no number *parser*. `Intl.NumberFormat.prototype` exposes `format`, `formatToParts`,
`formatRange`, `formatRangeToParts`, `resolvedOptions` — and nothing that reads a string back.
This is also why wger's version looks simpler than ours: Dart's `intl` ships
`NumberFormat.tryParse`, so their widget formats and parses from one object. We get the first
and not the second.

So the custom rule stays, but it is not invented: it is
[wger's](https://github.com/wger-project/flutter/blob/master/lib/core/number_input.dart)
rule, from the fix for [the same report against their app](https://github.com/wger-project/flutter/issues/1147),
and this PR already dropped the one place where I had gone beyond them.

Worth revisiting if we ever add `@formatjs/intl-numberformat` for another reason — at that
point react-aria's parser becomes viable and this function should be deleted in its favour.
I will add that as a note on the function so the next person doesn't have to redo this
research.

PR #150, opened 2026-08-26
Make every request settle, and give both apps one way to fail (#145)

### LINKED ISSUE #145: Action buttons stop responding / executing actions until app restart
**What happened?**
Action buttons stop triggering their intended actions (e.g., logging a set, saving changes, or completing an action). The UI remains rendered, but pressing the action buttons produces no response, requiring a full app restart to restore functionality.

**Steps to reproduce**
Currently unknown/undeterministic, occurs during normal usage.

**Expected behavior**
Action buttons should consistently register presses.

**Environment**
- Deployment: Docker
- Browser: mobile
- OS: Android 17 (CP2A.260805.005)

**Video file**
https://github.com/user-attachments/assets/4cc04188-1d7b-4b55-9d8e-f4963e0e7911

Cawlumm:
Fixes #145.

**Two independent defects were hiding behind one report.** The first came out of
@Cabiny6's video; the second only came out after they ran a build of that fix and told us
it still froze. The second one is the one that was actually biting them.

---

## The freeze: a promise that never settled

Their follow-up gave the two clues that name it — it happened when connectivity to their
homelab dropped, and it cleared *"immediately"* once the connection came back. Nothing was
looping. One promise had simply never settled.

Every request in the app goes through an axios instance with a 20s timeout. The token
refresh does not. It is issued on the bare `axios` export — using `api` would recurse into
the interceptor that issues it — and a bare axios call inherits axios's own default of
`timeout: 0`, meaning wait forever:

```ts
const res = await axios.post(`${base}/auth/refresh`, { refresh_token: refreshToken })
```

So a request that 401s while the network has gone quiet leaves its caller pending for the
life of the process. `saving` never clears, the Finish button stays disabled, and the app
carries on rendering at 60fps around it — the workout timer ticking beside a dead button in
their video, and the force-quit it took to clear.

Their access token lives an hour. A gym session is about an hour. So the request most
likely to go through the refresh path is the one that finishes the workout — *"almost
consistently when I start a workout and try to discard it or finish it."*

### Reproduced

A proxy that forwards everything except `POST /auth/refresh`, which it accepts and never
answers — gym wifi, faithfully modelled. A *stopped* server would have hidden this behind
an instant `ECONNREFUSED`; the black hole is what turned an intermittent report into a
one-tap repro.

```
  -> POST /api/v1/workouts
  <- 401 /api/v1/workouts
  BLACKHOLE POST /api/v1/auth/refresh
```

Finish sat on "Saving…" through six taps and two minutes, timer ticking. Restoring
connectivity then dropped the live workout straight to the login screen — losing it.

### The fix

Three changes, all in `packages/shared/src/client.ts`, so **web gets them too**:

**The refresh carries the same bound as everything else**, so every promise settles.

**A refresh nobody answered no longer ends the session.** Only the server can say a session
is over, and only 400/401/403 are it saying so. A timeout, a dropped connection or a 5xx is
silence, and silence is not a verdict — signing out on it ends a live workout every time
someone walks past a dead spot in their gym's wifi, which is exactly who self-hosts this.

**Concurrent 401s share one refresh.** A screen that mounts and fires five reads used to
open five, racing to rotate a token the server invalidates as it reissues it, so some
retries went out already dead — and on a bad network that is five separate 20s waits.

### What the research said

| source | |
|---|---|
| [wger's Flutter client](https://github.com/wger-project/flutter/blob/master/lib/core/network/auth_notifier.dart) | *"pure network errors keep the session intact so offline use continues to work"* — and `_refreshInFlight ??= …` for single-flight |
| [TanStack Query](https://tanstack.com/query/v4/docs/framework/react/guides/query-retries) | don't retry 401/403; handle them in a global handler. Pause transient failures instead |
| [Offline-first auth guidance](https://cursa.app/en/page/authentication-token-refresh-and-offline-authorization-decisions) | transient failure → *"keep the user signed in locally"*; invalid refresh token → require reauth |
| [axios ecosystem](https://gist.github.com/mkjiau/650013a99c341c9f23ca00ccb213db1c) | `isRefreshing` flag + queue, so N concurrent 401s make one refresh |

One deliberate divergence: wger clears the session on **any** non-200 refresh response. We
keep it on 5xx, because both these apps talk to a box the user owns — `docker compose up -d`
means a proxy answering 502 for a few seconds, and that must not evict anyone mid-workout.
A genuinely dead refresh token still 401s, so nothing hangs.

### Around the edges

`ConfirmSheet` takes an `error` and stays open, so a failed finish says why with the retry
under the same finger rather than dismissing to a banner at the top of a scrolled list. Gym
mode's finish swallowed its error entirely — a failure was indistinguishable from a tap that
never registered. Both screens now go through `apiErrorMessage`, since the failures that
matter here carry no response for a raw `err.response.data.error` to find. And
`classifyNetworkError` learned the shape Android actually produces for a timeout: the bare
string on the XHR, with no `ECONNABORTED` and no `_timedOut`.

---

## The other defect: two actors navigating on auth state

This is what the video showed, and it is real — just not what was freezing them.

On a failed refresh the API client cleared the stored tokens and pushed `/login`, but never
touched the auth store, so `isAuthenticated` stayed `true`. The root gate then saw an authed
user sitting in the `(auth)` group and pushed them back:

```ts
if (!isAuthed && !inAuthGroup) router.replace('/login')
else if (isAuthed && inAuthGroup) router.replace('/')   // ← straight back
```

Until React gave up with **`Maximum update depth exceeded`**. Reproduced by restarting the
backend with a different `JWT_SECRET` under a running app.

Access is now *declared* rather than navigated:

```tsx
<Stack.Protected guard={isAuthed}><Stack.Screen name="(tabs)" /></Stack.Protected>
<Stack.Protected guard={!isAuthed}><Stack.Screen name="(auth)" /></Stack.Protected>
```

Complementary guards, so exactly one group is reachable at any moment and there is no
navigation call left to argue with the store. This is what both frameworks prescribe —
[React Navigation](https://reactnavigation.org/docs/auth-flow/): *"you **don't manually
navigate**"*; [Expo](https://expo.dev/blog/simplifying-auth-flows-with-protected-routes):
`Stack.Protected` exists to *"avoid imperative redirects"*. It shipped in Expo Router v5;
this app is on v6 and was still redirecting by hand.

An earlier version of that fix added a fixed-point redirect helper and a circuit breaker
that force-signed-out after four identical redirects in a second. Both are deleted — needing
a defence like that was the signal that the pattern itself was wrong.
`authNavigation.guard.test.ts` keeps it that way: it fails if anything outside the root
layout routes to `/login`, in any of the three ways expo-router offers.

Which screen a signed-out user lands on is now declared too. `sortRoutes` breaks its final
tie on route-name **length**, not alphabetically, so `login` only beat `register` by being
shorter; `unstable_settings.anchor` pins it.

---

## Verified

Android emulator, Expo Go.

| | |
|---|---|
| finish with the refresh black-holed | ✅ error in the sheet, session kept, workout kept, button live |
| …the message it shows | ✅ "Can't reach the server… (timeout)" — not "invalid or expired token" |
| cold start with the network black-holed | ✅ stays signed in, empty states, no hang |
| finish on a long-expired token, network restored | ✅ one refresh for the whole dashboard burst, workout saved |
| session killed under a running app | ✅ clean drop to login, zero update-depth errors |
| deep navigation, back button, repeated sign-in/out | ✅ |

Every new test was checked against the old code first — the five that should fail there do.

| gate | |
|---|---|
| shared | 266 across 19 suites |
| mobile | 43 across 10 suites |
| type-check | clean (shared, mobile, web) |
| lint | 0 errors |

---

## Taking the lesson across both apps

#145 was one call site getting one thing wrong. It was worth asking how many
other call sites were positioned to get the same thing wrong, and the answer was
most of them.

**Forty-odd copies of the same six lines** — set a busy flag, try, await, catch, set a
message, clear the flag — that did not agree with each other:

| what they did | how many | what the user saw |
|---|---|---|
| read `err.response.data.error` directly | 12 | "Failed to save" on any timeout — the failure that carries no response |
| `catch {}`, then quietly restore the row | 9 | nothing at all; identical to a tap that missed |
| set the message, then dismiss the sheet showing it | 2 | nothing at all |
| report `err.message` | 3 | "Request failed with status code 400" |

**`useAsyncAction`** (`@lyftr/shared`) now owns that shape, with `apiErrorMessage`
**inside** it. The wording is no longer something a call site can bypass, forget or
half-remember — which is the only version of "one place" that survives contact with the
next feature. 38 call sites across both apps.

It deliberately does not decide what success means: navigating, closing a sheet,
refetching a list stays in the action, where the screen can see it. Where a handler
validates a value before sending, that value is passed to `run()` — the guard in the
caller is what narrows the type.

**Web gains the `ConfirmSheet` mobile already had.** Seven pages hand-rolled the same
portal and had drifted apart; none had anywhere to put an error, so a confirm that failed
just closed. `DiscardConfirm` was an earlier, narrower run at the same de-duplication and
is deleted. Zero hand-rolled confirm portals remain.

### What is deliberately not converted

Ten sites keep their own busy flag, because they are a different process:

- **Loads** (`ExercisePicker` on both, `Food`, `WorkoutExercisePicker`, `workouts/start`) —
  they populate a screen rather than commit a change, and carry request-sequence guards
  and pagination the hook has no business knowing about.
- **Probes** (the Server URL test on sign-in and settings) — `testServerConnection`
  returns `{ok, message}` and never throws. There is nothing to catch.
- **Auth** (`Login`, `Register`) — the auth store already owns `isLoading` and `error` and
  already routes through `apiErrorMessage`. A second owner is the bug this branch is about.

### Where each thing lives now

| process | one place |
|---|---|
| error → words | `apiErrorMessage` — 0 call sites read `err.response.data.error` |
| run a request that can fail | `useAsyncAction` — 38 call sites |
| confirm a destructive action | `ConfirmSheet` — web 5, mobile 11 |
| render a message | web `.alert-*`, mobile `<Alert>` |

## Follow-ups, not in this PR

- **Pre-emptive refresh.** wger refreshes when the token is within 30s of expiry and calls
  that its *primary* safeguard — it avoids the 401 dance while the user still has signal.
  Needs client-side JWT `exp` decoding.
- **Say why you were signed out.** wger carries a `sessionExpired` flag and hints on the
  login screen. Both our apps drop to login with no explanation. Web also still does
  `window.location.href = '/login'` rather than reacting to store state.


backend/utils/validation.go
```diff
@@ -0,0 +1,145 @@
+package utils
```
Cawlumm, line 1: instead of in this file would we be able to use something like https://github.com/go-playground/validator, or similar? Stick to go standards.

packages/shared/src/hooks/useAsyncAction.ts
```diff
@@ -0,0 +1,77 @@
+import { useCallback, useEffect, useRef, useState } from 'react'
```
Cawlumm, line 1: Wonder if this is worth doing for our global errors to bubble up instead of what we are currently doing. To bubble up global errors (like a 401 Unauthorized token expiry, a network outage, or a server 503 error) without forcing your individual screens to handle them, you can pair your hook with a Global Error Event Bus or a React Context Queue.
Since your useAsyncAction hook already centralizes error handling inside apiErrorMessage, we can intercept critical global errors inside that function (or right next to it) and push them to a layout-level banner, leaving the component to only manage its local operational state.
Here is the best-practice design pattern to achieve this without complicating your hook's API.
------------------------------
## 1. Create a Global Notification Queue (Context + Hook)
First, we build a simple, centralized system to hold and display global errors. This component sits at the root of your application layout (e.g., above your router).

import React, { createContext, useContext, useState, useCallback } from 'react'
export interface GlobalError {
  id: string
  message: string
}
const GlobalErrorContext = createContext<((message: string) => void) | null>(null)
export function GlobalErrorProvider({ children }: { children: React.ReactNode }) {
  const [errors, setErrors] = useState<GlobalError[]>([])

  // Pushes an error to the layout-level queue
  const addGlobalError = useCallback((message: string) => {
    const id = Math.random().toString(36).substring(2)
    setErrors((prev) => [...prev, { id, message }])
    
    // Auto-dismiss after 5 seconds
    setTimeout(() => {
      setErrors((prev) => prev.filter((e) => e.id !== id))
    }, 5000)
  }, [])

  return (
    <GlobalErrorContext.Provider value={addGlobalError}>
      {children}
      {/* Global Banner Overlay */}
      <div className="global-error-toast-container">
        {errors.map((err) => (
          <div key={err.id} className="error-toast">
            {err.message}
          </div>
        ))}
      </div>
    </GlobalErrorContext.Provider>
  )
}
// Low-overhead hook for internal interceptors to dispatch messagesexport function useAddGlobalError() {
  const context = useContext(GlobalErrorContext)
  if (!context) throw new Error('useAddGlobalError must be used within GlobalErrorProvider')
  return context
}

------------------------------
## 2. Standardize apiErrorMessage to Detect Global Failures
Your apiErrorMessage helper needs to inspect the error object, determine if it requires a global system alert or a local form alert, and return an interception strategy.

// client/errors.ts
export class GlobalInterceptError extends Error {
  constructor(message: string, public isAuthFailure = false) {
    super(message)
  }
}
export function apiErrorMessage(err: unknown, fallback: string): string {
  // 1. Handle network timeouts or drops
  if (!navigator.onLine || (err instanceof Error && err.message.includes('timeout'))) {
    throw new GlobalInterceptError('Your internet connection appears offline. Please try again.')
  }

  // 2. Mock Axios/Fetch response checking
  const response = (err as any)?.response
  if (response) {
    // 401 Unauthorized -> Needs global redirect/eviction notice
    if (response.status === 401) {
      throw new GlobalInterceptError('Your session has expired. Please log in again.', true)
    }
    // 502/503/504 Bad Gateway / Maintenance
    if (response.status >= 500) {
      throw new GlobalInterceptError('The server is experiencing technical difficulties. Please try again later.')
    }
    // 3. Expected API validation error (return locally to the form field)
    return response.data?.error || fallback
  }

  return fallback
}

------------------------------
## 3. Integrating the Interceptor into useAsyncAction
Now, we slightly modify your useAsyncAction catch block. If apiErrorMessage throws a GlobalInterceptError, the hook interceptor catches it, passes it to the global queue banner, and keeps the local screen clean.

import { useAddGlobalError } from './GlobalErrorContext'import { GlobalInterceptError, apiErrorMessage } from '../client'
// ... Keep your same interfaces and setup ...
export function useAsyncAction<A extends unknown[] = []>(
  action: (...args: A) => Promise<unknown>,
  fallbackMessage: string,
): AsyncAction<A> {
  const [busy, setBusy] = useState(false)
  const [error, setError] = useState('')
  const addGlobalError = useAddGlobalError() // 👈 Access the global channel

  const latest = useRef(action)
  latest.current = action
  const fallback = useRef(fallbackMessage)
  fallback.current = fallbackMessage

  const mounted = useRef(true)
  // ... Keep your exact useEffect mounted tracking ...

  const run = useCallback(async (...args: A): Promise<boolean> => {
    setBusy(true)
    setError('')
    try {
      await latest.current(...args)
      return true
    } catch (err) {
      if (!mounted.current) return false

      try {
        // Enforce parsing. If it's a global breakdown, it throws out of this try block.
        const localMessage = apiErrorMessage(err, fallback.current)
        setError(localMessage)
      } catch (intercepted) {
        if (intercepted instanceof GlobalInterceptError) {
          // Push the message to the top-level layout queue
          addGlobalError(intercepted.message)
          
          if (intercepted.isAuthFailure) {
            // Option: trigger global logouts, clear local storage tokens, 
            // or window.location.href = '/login'
          }
        } else {
          // Fallback safeguard
          setError(fallback.current)
        }
      }
      return false
    } finally {
      if (mounted.current) setBusy(false)
    }
  }, [addGlobalError]) // 👈 Safely added dependency (stable from context)

  const reset = useCallback(() => setError(''), [])

  return { busy, error, run, reset }
}

## Why this extension maintains your architecture's integrity:

   1. Zero Call-Site Changes: Your standard forms (const save = useAsyncAction(...)) require absolutely zero code edits. They continue reading {save.error} for inline validation anomalies (like "Email already taken").
   2. Context Stability: The context provider passes down a raw setErrors state dispatch wrapped in a memoized useCallback. This guarantees that addGlobalError never changes identity, leaving your hook's core feature (stable function identity for run) fully intact.

Would you like to explore how to implement an Axios Interceptor or native fetch wrapper instead, which catches these errors globally before they even touch your component lifecycles?

Cawlumm:
Closing this unmerged: everything in it has shipped, as smaller PRs that were each reviewed and manually tested on their own.

| What | Where it landed |
|---|---|
| Bound the token refresh; silence is not a verdict (the #145 hang itself) | #153 |
| The shared client every failing request goes through | #158 |
| `useAsyncAction`, `useServerList`, `apiErrorMessage`, `isNotFound` and the rest of the shared plumbing | #158, #161 |
| Web: a failed load says it failed — page / section / stat-tile errors, no invented zeros | #163 |
| `REVIEW.md` | #175 |
| Go tests that rotted with the calendar | #176 |
| Mobile: a failed save or delete says so, where the tap was | #177 |
| Mobile: semantic colour is the theme's decision, held to AA | #178 |
| Mobile: pickers on `useServerList`, one scanner, one inline confirm | #179 |
| Mobile: a failed load says it failed, scoped to what failed | #180 |

Main is green after each. This branch is left in place rather than deleted, in case anyone wants to diff against the original.

Two things this work turned up that are deliberately **not** in any of the above, and want their own issues rather than being buried here:

- Calorie and macro targets are still invented when `/settings` fails to load. The shared settings store already exposes a `loadFailed` flag for exactly this, and only the mobile Settings screen reads it — so "/ 2000 kcal" and "1,340 left" can be defaults the user never set, on both apps.
- On Weight, tapping a period fires the trend request twice and the extra reload discards every loaded page of history; `reload()` is also a silent no-op while another page is in flight, so a delete during pagination leaves the deleted row on screen.

Issue #41, opened 2026-06-26
Input food by weight


valterschutz:
**What problem does this solve?**
I don't always eat food in multiples of .5 servings. Having the ability to enter the exact weight in terms of grams would be very useful.

**Describe the feature**
After selecting food item, toggle between input in terms of servings and in terms of grams.

**Who benefits?**
Anyone who takes food tracking seriously.

PR #186, opened 2026-09-20
Log food by weight, and read the pack's own serving

### LINKED ISSUE #171: Some Products Information is Just Plain Wrong
I have noticed that a number of the foods you can access via scanning a barcode just have very wrong data. This ranges from eggs to olive oil to more.

The best example I've found (that inspired me to finally raise this issue) is Thrive Market olive oil.

The actual product:
Image

How it looks in Lyftr:
Image

According to Lyftr, one serving is 800 calories and 93.3g of fat. In reality, each serving is only 120 calories and 14g of fat. That's a massive difference.

Where does Lyftr pull this data from? Is it stored independently on the server and get refreshed per version update or does the server have to reach out someplace to get the info in real time? And is there any way to contribute to more accurate data?

Thanks!

Cawlumm:
Fixes #171 and #41.

## What was wrong

@pdschneider scanned a bottle of Thrive Market olive oil and got **800 kcal / 93.3 g fat for "1 serving"**, against a label reading 120 kcal / 14 g. @ansgaross worked out why: those are per-100, and the pack's serving is 15 ml.

Three separate defects sit behind that, and the numbers themselves are not one of them.

**1. Per-100 ml figures labelled "100 g".** OpenFoodFacts has no per-serving data for that product — in a sampled 20 oils, half have none — so we fall back to its per-100g values, which for a liquid *are* per 100 ml. The fallback now says "100 ml" when the serving is a volume. The unit is read off the label text, not OpenFoodFacts' own `serving_quantity_unit`, which reports `ml` for every cup-based serving (`"1/4 cup (7 g)"` comes back as ml).

**2. 15 ml was unloggable.** Both clients floored the amount at 0.5 servings, so against a 100 ml basis the least anyone could record was 50 ml — *"an unusable item because of the way it calculates servings"*. The server never had a floor. The field now takes grams or millilitres and derives the servings the diary stores, falling back to servings — unfloored — when nothing knows what a serving measures. That is #41's ask.

**3. Search and a scan disagreed.** The OpenFoodFacts search index returns only `*_100g` and omits `serving_size` entirely, so the same Market Pantry olive oil read `800 kcal / "100 g"` from a search and `120 kcal / "1 Tbsp (15 ml)"` from a scan. Selecting a search hit now re-reads the product through the barcode endpoint first. If that read fails, the search figures still log, under a line saying which figures they are.

## How it was decided

wger, Waistline, OpenNutriTracker and FitBook all treat per-100g as canonical and enter amounts in grams; MacroFactor and YAZIO make "per 100 g / per 100 ml" an explicit basis and let you type the unit. None of them floors the amount. Two references worth keeping: [openfoodfacts-server#7768](https://github.com/openfoodfacts/openfoodfacts-server/issues/7768) documents 7,700+ products where `serving_quantity` is null despite a serving being set — which is why `0` stays meaningful here rather than defaulting to 100 — and [SparkyFitness#2416](https://github.com/CodeWithCJ/SparkyFitness/issues/2416) is this exact bug in another OpenFoodFacts tracker.

## Shape

| Layer | Change |
|---|---|
| `controllers/food.go` | parse `serving_quantity` (number *or* numeric string); emit `serving_quantity` + `serving_unit`; "100 ml" for liquids; carry the barcode on search hits |
| `db/migrations.go` | append `serving_quantity` / `serving_unit` to `food_logs` and `saved_foods`; 0 means "unknown", so nothing is backfilled or guessed |
| `shared/hooks/useFoodAmount.ts` | **new** — the amount field's arithmetic, one copy for both apps |
| `shared/utils/food.ts` | `servingBasis`, `servingsForAmount`, `amountForServings`, `formatServings`, `formatLoggedAmount` |
| web + mobile | amount field with a unit suffix and a servings readout; the 0.5 floor deleted; search hits re-read; the diary reads back "15 ml", not "0.15 servings" |

## Verified

- Go `./...`, `go vet`, gofmt (the two files I touched were already unformatted on main; formatting them is in the diff)
- shared 376, web 130, mobile 65, Playwright `food.spec` 33/33 — all green
- Against the **live** OpenFoodFacts API on a real backend: an olive oil with serving data reads 120 kcal per 1 Tbsp (15 ml); one without reads per **100 ml** and logs 15 ml as 124 kcal / 13.8 g fat
- The migration ran clean on a dev database that already held food logs

**Not verified on a device.** The mobile change is byte-for-byte the same logic as web via the shared hook, and type-checks and jest pass, but reaching the logged-in mobile screens needs credentials typed into the app, which I don't do. Worth a look on a device before release.

https://claude.ai/code/session_01GY7Lc9HhTWd1D65qC74jA4


mobile/app/(tabs)/nutrition/log.tsx
Cawlumm, line 126: Comment view
  }, [])
Check warning on line R126
Check warning:
React Hook useCallback has missing dependencies: 'listEpoch' and 'setSavedFoods'. Either include them or remove the dependency array
CI / Mobile + shared checks
Check warning on line R126
Check warning:
React Hook useCallback has missing dependencies: 'listEpoch' and 'setSavedFoods'. Either include them or remove the dependency array

